PR #7, opened 2022-11-22
Fix typo (used Typescript instead of Typeshare)

julienfouilhe:
I was reading the documentation for this interesting project and found a small error!

Thanks for sharing!


InquisitivePenguin:
Great catch! Thanks 😄

PR #9, opened 2022-11-22
Fix README typo Typescript->Typeshare

drrlvn:


drrlvn:
Oops just noticed a PR for this already exists, so closing. Sorry for the noise!

PR #16, opened 2022-11-23
Extract format so it can be changed per language

exoego:
Right now, all supported languages format generic parameters into `<A, B, C>`.
By this change, we can introduce different format, like `[A, B, C]` for Scala.


exoego:
Superseded by https://github.com/1Password/typeshare/pull/27

PR #19, opened 2022-11-23
fixed: arguments in docs

ChrisMcKenzie:
the readme seems to use an outdated argument this just updates it


CerulanLumina:
Thanks for catching this!

PR #27, opened 2022-12-05
Add Scala support

exoego:
This PR adds Scala language support that passes all existing test cases.
As far as I tested, the generated codes support both Scala 2 and Scala 3.

Some notes on design:

- Scala does not support unsigned integers. But for many cases, the signed integers can fit. Java 8+ offers [operations to emulate unsigned, such as Long.compareUnsigned](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#compareUnsigned-long-long-). So my impl converts unsigned ones to signed ones, with type aliases such as `type UInt = Int` to indicate it is unsigned. It is better than "unsigned integer is not supported" panic.
- Scala 2 does not allow type aliases defined in top-level. Since Scala 2 is still the majority, type aliases are put in `package object`. 
- Extracted `fn format_generic_parameters` since `T<A, B>` is the format of generic type for other supported languages but `T[A, B]` for Scala.
- Since Scala has no built-in support of renaming members on serialization or de-facto standard for it, raw names are encoded into `serialName: String` members. Users should use `serialName` on serializers.
- To generate the code in a single file, either or both of `package object FOO {}` and `package FOO {}` may appear in the generated file. See the below example for explanation:

```scala
package com.example // Scala allows multiple package statements in a single file

package object typeshare {  // Which means com.example.typeshare
  type UByte = Byte
  // type alias goes here since type aliases should be defined in class or (package) object in Scala 2.
}

package typeshare {  // Which also means com.example.typeshare
  sealed trait MyEnum {
    def serialName: String
  }  
  object MyEnum {
    case class Foo(age: UByte) extends MyEnum {
      val serialName: String = "foo"
   }
    case object Bar extends MyEnum {
      val serialName: String = "b_a_r"
    }
  }
}
```


InquisitivePenguin (approved):

InquisitivePenguin:
@exoego Thank you for all the work you've done on this!

PR #28, opened 2022-12-05
Rename typeshare-cli binary to `typeshare` on installation

### LINKED ISSUE #4: Inconsistant CLI binary name usage
## What

Currently the majority of the documentation references installing the `typeshare-cli` but then running `typeshare` to invoke it why currently does not work out-the-box unless you alias `typeshare` to `typeshare-cli`.

## Possible fix

Either update the documentation to use `typeshare-cli` everywhere, rename the CLI binary from `typeshare-cli` to just `typeshare` or offer some kind of shortcut / alias for the binary out-the-box (Not looked into Cargo binary distribution in a while so can't remember / don't know if this is possible)

InquisitivePenguin:
This fixes #4.

Rather simple to solve, just had to add a `[[bin]]` entry to `cli/Cargo.toml`.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -8,6 +8,10 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/1Password/typeshare"
 readme = "README.md"
 
+[[bin]]
+name = "typeshare"
+path = "src/main.rs"
+
 [features]
 go = []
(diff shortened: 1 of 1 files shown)


CerulanLumina (approved):
Simple clean fix, I like it.

PR #37, opened 2022-12-18
Updates path to contributing page

kareid:
The usage.md page link to contribution information was leading to a 404 instead of a valid page.  This change corrects the path and page name.


CerulanLumina (approved):
:+1: 

Now to figure out why the pipeline is failing...

PR #45, opened 2023-01-01
Fix tests when go is not enabled

ccouzens:
```bash
cargo test
cargo test --features go
```

The struct is defined with the `cfg` macro on the field, so makes sense to test it the same way.

```rust
/// The paramters that are used to configure the behaviour of typeshare
/// from the configuration file `typeshare.toml`
pub(crate) struct Config {
    pub swift: SwiftParams,
    pub typescript: TypeScriptParams,
    pub kotlin: KotlinParams,
    #[cfg(feature = "go")]
    pub go: GoParams,
}
```


CerulanLumina (approved):
Looks good, thank you for your contribution!

InquisitivePenguin (approved):

PR #51, opened 2023-01-11
Support unit structs that don't use bracket syntax

### LINKED ISSUE #41: Support for unit structs
Hi,

Currently, unit structs are not supported:

```rust
#[typeshare::typeshare]
pub struct MyType;
```

(Error: `unit structs are currently unsupported`)

A workaround is to create a tuple struct with a tuple as argument (tuple structs without an argument result in another error):

```rust
#[typeshare::typeshare]
pub struct MyType(pub ());
```

However, this is ugly and results in the following code being generated (Typescript):

```typescript
export type MyType = undefined;
```

I'm a Typescript beginner, but this doesn't look great. I think, it would be better to have something like:

```typescript
export type MyType = "MyType";
```

(I'm not saying that `struct MyType(());` should result in the above, but that `struct MyType;` should be supported, which would generate it).

InquisitivePenguin:
This PR will add support for unit structs that use semicolon syntax instead of bracket syntax.

In other words,
```rust
#[typeshare]
struct UnitStruct;
```
will now work properly.

This closes #41


core/data/tests/can_generate_unit_structs/output.kt
```diff
@@ -0,0 +1,14 @@
...
+
+@file:NoLiveLiterals
+
+package com.agilebits.onepassword
+
+import androidx.compose.runtime.NoLiveLiterals
+import kotlinx.serialization.*
+
+@Serializable
+class UnitStruct
```
McAJBen, line 13: @InquisitivePenguin 
I would expect this to use Kotlin's `object` syntax like this repo does for enums. This would make it similar to tests `boxed_value` and `can_handle_anonymous_struct`.
```kt
@Serializable
@SerialName("UnitStruct")
object UnitStruct
```

InquisitivePenguin: Thanks for bringing this up. I can open a PR to follow-up on this once this PR is merged.

CerulanLumina (approved):
Looks good

Issue #21, opened 2022-11-24
Type alias with option outputs invalid TS


azdavis:
Given this Rust input:

```rs
#[typeshare]
pub type Foo = Option<u32>;
```

The TypeScript output is:

```ts
export type Foo? = number;
```

This is invalid syntax.

InquisitivePenguin:
Thank you for opening this issue. I'll look into why this is happening.

InquisitivePenguin:
Fixed in #52. @azdavis or anyone else feel free to open this issue if it happens again.

PR #52, opened 2023-01-11
Fix typescript optional type alias formatting

McAJBen:
Typescript only allows ? in properties and arguments, not in type aliases.
In type aliases I've replaced it with " | undefined".

This is a potential fix for https://github.com/1Password/typeshare/issues/21


InquisitivePenguin (approved):
Thank you for fixing this!

PR #25, opened 2022-11-27
First pass at Python implementation

adriangb:
Todo:
- A lot of cleanup, this is pretty hacky
- Fix line spacing
- ~Make ordering of globals deterministic (the generated globals are not consistently ordered, which obviously makes tests fail)~ Done
- Consider vendoring / implementing a topological sort

PR #53, opened 2023-01-12
Make the Language trait mutable

adriangb:


adriangb:
Side note: you might want to change the Actions security settings:

212133196-5ae743e2-e888-44ae-bdbd-7dfba49fcb9d

InquisitivePenguin:
@adriangb Is this a prerequisite for another change you plan on making, or is this meant to be a general improvement that will enable more flexibility for future implementations?

adriangb:
A bit of both. I first encountered the need for this in #25, but I think it might be useful for other implementations as well. Python needs it to keep track of imports, I implemented that in #25 with a RefCell but it would be a lot nicer if Language was mutable. I thought it would be a simple change but unfortunately, it broke things in other implementations because of borrowing closures, so it may not be that simple.

InquisitivePenguin:
I see! I can help fix the errors here if you want.

InquisitivePenguin:
I've fixed your errors in this PR here: https://github.com/adriangb/typeshare/pull/1

adriangb:
Awesome! Thanks for doing all of the hard work 😄

InquisitivePenguin (approved):

PR #60, opened 2023-02-09
Allow to define `readonly` TypeScript properties

### LINKED ISSUE #50: Making readonly properties in TypeScript
Hello,
Is there any possibility to make force a generated TypeScript interface to have a readonly modifier on a property?

Czocher:
This PR adds a `typescript(readonly)` decorator which allows for the creation of `readonly` fields in TypeScript.

Closes #50.


core/src/parser.rs
```diff
@@ -460,6 +464,12 @@ fn get_field_type_override(attrs: &[syn::Attribute]) -> Option<String> {
     attr_value("typeshare", attrs, PREFIX, SUFFIX)
 }
 
+fn get_field_readonly(attrs: &[syn::Attribute]) -> bool {
+    const PREFIX: &str = "typescript";
+    const SUFFIX: &str = "(readonly)";
+    attr_value("typeshare", attrs, PREFIX, SUFFIX).is_some()
+}
```
InquisitivePenguin, line 471: Ideally we should have a more extendable way to handle language-specific parameters like this. If we ever add future configuration here, this code will break.

This ties back into #63, which is also about implementing language-specific features. I was going to do some refactoring of our attribute code to support this, so maybe we could combine our work here?

InquisitivePenguin (changes requested):

core/src/parser.rs
Czocher: Yes, I noticed it's rather hack-ish when I implemented it and hoped for some ideas during review. Thank you @InquisitivePenguin for sharing. I'll gladly rebase this PR on top of a more generic solution.

InquisitivePenguin: Thanks! Work to update the parser is happening in https://github.com/1Password/typeshare/pull/66

InquisitivePenguin: The parser work has been merged into main, so you should be able to update this PR accordingly once you pull from main.

Czocher: Thanks for the notification @InquisitivePenguin!

Czocher:
Hello @InquisitivePenguin,
I implemented the necessary changes and generalized it over all supported languages.
Currently it parses every `#![typeshare({language}(some, decorators, here))` annotation and provided the defined annotations for further analysis.

InquisitivePenguin:
> Hello @InquisitivePenguin, I implemented the necessary changes and generalized it over all supported languages. Currently it parses every `#![typeshare({language}(some, decorators, here))` annotation and provides the defined decorators for further use.

Thanks! Sorry for the delay on reviewing this.

Czocher:
> > Hello @InquisitivePenguin, I implemented the necessary changes and generalized it over all supported languages. Currently it parses every `#![typeshare({language}(some, decorators, here))` annotation and provides the defined decorators for further use.
> 
> Thanks! Sorry for the delay on reviewing this.

No problem, thanks for your hard work!

InquisitivePenguin (approved):
Thanks for getting this done!

PR #61, opened 2023-02-09
Support the double option pattern for TypeScript

### LINKED ISSUE #59: Double option pattern support
Hello,
I found a missing feature in `typeshare`. Some APIs support partial updates for DTOs. That means that they have a distinction between a missing field and a field equal to `null` eg.:

```rust
#[derive(Serialize, Deserialize, Debug)]
#[typeshare]
pub struct UserUpdateParams {
    pub username: Option<String>,
    pub password: Option<String>,
    #[typeshare(serialized_as = "Option<String>")]
    pub valid_until: Option<OffsetDateTime>,
}
```

When the `valid_until` property is `null` on the database level, it means that the user is valid forever. To support this kind of use case `serde` has a [`double option`](https://docs.rs/serde_with/latest/serde_with/rust/double_option/) pattern.

Unfortunately this is not supported by `typeshare` since `Option<Option<String>>` is compiled as `valid_until?: string` in `TypeScript`. Instead it should be compiled as `valid_until?: string | null`.

Czocher:
Double option pattern is used for partial updates of optional fields. See the [serde_with helper](https://docs.rs/serde_with/latest/serde_with/rust/double_option/) for more details.

Closes #59.


core/src/language/typescript.rs
```diff
@@ -219,12 +219,14 @@ impl TypeScript {
...
+        let double_optional = field.ty.is_double_optional();
         writeln!(
             w,
-            "\t{}{}: {};",
+            "\t{}{}: {}{};",
             typescript_property_aware_rename(&field.id.renamed),
             optional.then(|| "?").unwrap_or_default(),
-            ts_ty
+            ts_ty,
+            double_optional.then(|| " | null").unwrap_or_default()
```
InquisitivePenguin, line 229: I'm not a typescript expert - is there a reason why this was chosen as `null` instead of `undefined`?

Czocher: Sure, there's actually a difference here. `undefined` usually means that a given field does not have a defined value, `null` specifically means that the field does have a value, but it's a `nothing` value.

In this context `null` is the right choice, since we specifically want to declare that we want to "zeroize/reset" the existing value during a partial update.

InquisitivePenguin: Thanks for the explanation!

InquisitivePenguin (approved):

Issue #20, opened 2022-11-24
Support Python with pydantic


higumachan:
I regularly exchange data between Rust and Python (with pydantic).
When I saw this project, I really wanted to be able to incorporate it into my workload.

Would I be welcome to make python-enabled contributions?

Also, it would be helpful if you could share any concerns you have with python support.

adriangb:
Started giving this a shot. Here are a couple initial thoughts:

## Order of globals matters
```rust
type MyTypeAlias = MyType
struct MyType{}
```

By default this library writes out globals in a fixed order: https://github.com/1Password/typeshare/blob/e461bb680bb1b0f943bee6c38498eaf1027f62e9/core/src/language/mod.rs#L30-L40

The Python implementation will need to override this and topologically sort globals to write them out in a valid order. I was generally able to do this but:

- It's pretty invasive overriding of `generate_types`
- We'd have to either bring in a dependency or vendor a topological sorting algorithm
- We need to error out if there are cycles (Python doesn't support this)

I'd consider changing the default to topological sorting since I don't think that will be a problem for any other language and is likely more readable.

## Mutable state on `Language`

Python needs some sort of mutable state to keep track of the imports it needs (`from pydantic import BaseModel`, `from typing import Dict`, etc.)

Each `Language` is supposed to be able to have mutable state:

https://github.com/1Password/typeshare/blob/main/core/src/language/mod.rs#L21

But the trait methods are declared with `&self` so `self` is immutable:

https://github.com/1Password/typeshare/blob/main/core/src/language/mod.rs#L27

I think it should have been `&mut self` but changing that would be a breaking change (it seems like Rust doesn't have the ability to "refine" traits to a more-general version w.r.t. mutability).

I got around this using `RefCell` but this can lead to runtime panics and makes things a lot more clunky.

## Python versions and `typing-extensions`

The best way to a unit enum  or a tagged union with a string discriminant

```rust
enum Foo {
  Bar,
  Baz,
}

#[typeshare]
#[serde(tag = "type", content = "content")]
pub enum EnumWithManyVariants {
    UnitVariant,
    TupleVariantString(String),
}
```

Is

```python
Foo = Literal["Bar", "Baz"]

class EnumWithManyVariantsUnitVariant(BaseModel):
  type = Literal["UnitVariant"]

class EnumWithManyVariantsTupleVariantString(BaseModel):
  type = Literal["TupleVariantString"]
  content: str

[...]

PR #57, opened 2023-01-27
`const` support

snowsignal:
Closes #44.

This PR will allow the `typeshare` annotation to be used for `const` variables. This will copy the variable declaration into the output for the respective language.

For example:
```rs
const ANSWER_TO_EVERYTHING: u32 = 42;
const BEST_FILM: &str = "The Tree of Life";
``` 

This would become:
```ts
const ANSWER_TO_EVERYTHING = 42;
const BEST_FILM = "The Tree of Life";
```
```swift
class Constants {
  static let ANSWER_TO_EVERYTHING = 42
  static let BEST_FILM = "The Tree of Life"
}
```
```kt
const val ANSWER_TO_EVERYTHING = 42
const val BEST_FILM = "The Tree of Life"
```
```go
const ANSWER_TO_EVERYTHING = 42
const BEST_FILM = "The Tree of LIfe"
```

As of right now, this PR only supports integers, floats, and strings as constant types.

PR #62, opened 2023-02-10
Write out types in topological order

adriangb:
Sub-piece of #25.

Python in particular needs this for type aliases:

```python
TypeAlias = Foo
class Foo:  pass
```

Won't work although it is fine in Rust.

As discussed in #20 I think it makes sense to just make this the default for all languages, I don't think it's a con in any language.

Tricky bits:
- Cycles. Python is fine with cycles as long as it's not in type aliases. For now I just let cycles be (no error). We can either leave it like that (which would result in generated Python files with cycles in type aliases that fail to import) or try to do something fancy for the Python implementation specifically that checks if the cycle includes a type alias and gives a nice error message.
- Duplication of `RustThing`. I essentially duplicated `RustThing` from `parser.rs` to `rust_types.rs` because it made sense for topological sorting. I think this needs some cleanup / further thought.
- Handling of languages like Go that override `write_types`. I copy pasted logic into the Go implementation (it already has duplicate logic w.r.t `mod.rs`) although it is not strictly necessary for Go, just to keep things the same across languages.


adriangb:
@InquisitivePenguin what would you think about making the RustThing in `parser.rs` public? Then we can add a `fn toposorted() -> Vec<RustThing>` method to ParsedData.

InquisitivePenguin:
> @InquisitivePenguin what would you think about making the RustThing in `parser.rs` public? Then we can add a `fn toposorted() -> Vec<RustThing>` method to ParsedData.

I'm hesitant to do so because `RustThing` is kind of a hack so if we changed this in the future we'd need to do a major version update because this would break backwards compatibility.

adriangb:
That’s fair. Do you consider it a hack just as it is currently being used in `parser.rs` or also how I am using it here to iterate over heterogeneous “things” in a topological order?

core/src/rust_types.rs
```diff
@@ -485,10 +485,17 @@ impl RustEnumVariant {
...
+#[derive(Debug, PartialEq)]
 pub struct RustEnumVariantShared {
     /// The variant's ident
     pub id: Id,
     /// Comments applied to the variant
     pub comments: Vec<String>,
 }
+
+#[derive(Debug, PartialEq)]
+pub enum RustThing<'a> {
```
InquisitivePenguin, line 497: We need an `#[allow(missing_docs)]` here. A brief description of why it needs to be duplicated would also be helpful.

InquisitivePenguin: Also I'd prefer if we keep things like this `pub(crate)`, since this is still technically a hack.

InquisitivePenguin:
> That’s fair. Do you consider it a hack just as it is currently being used in `parser.rs` or also how I am using it here to iterate over heterogeneous “things” in a topological order?

It was originally introduced to _accomplish_ a hack, in this case treating a struct as a type alias. I'm not opposed to have a heterogeneous container type, but since this is going to be a new public type, we could still make a few changes:

1) Rename it to something better (I don't really have any strong opinions on a new name)
2) Make it `#[non_exhaustive]` so future matching code won't break when we add new parsed data types (which #57 is already doing)
3) Figure out if this would work better as a trait (we'd have to downcast to get the data back, so probably not)

core/src/rust_types.rs
InquisitivePenguin: Alternatively, we could do what I mentioned in https://github.com/1Password/typeshare/pull/62#issuecomment-1426934009

InquisitivePenguin:
@adriangb Since this PR is getting kind of stale, I'm going to go ahead and implement the requested changes myself. Let me know if this is conflicting with any work you're doing.

adriangb:
No problem at all! I've been very distracted from OSS with personal / paid job lately, sorry for dropping the ball.

InquisitivePenguin:
@adriangb No worries at all, thank you for all of your contributions.

core/src/topsort.rs
```diff
@@ -0,0 +1,225 @@
...
+        graph,
+        &(0..graph.len()).collect(),
+        &mut res,
+        &mut processed,
+        &mut seen,
+    );
+    res
+}
+
+pub(crate) fn topsort<'a>(things: Vec<&'a RustItem>) -> Vec<&'a RustItem> {
```
CerulanLumina, line 183: Clippy wants elided lifetime here

CerulanLumina (approved):
Everything looks good, just a few nits:

- could the other functions taking mutable vecs in topsort.rs have shims like the topsort impl? It'd reduce extra lines needed around those functions. Doc comments on those would be nice as well. I don't think this is significant enough to block merge though.

adriangb:
Thank you for taking this over @InquisitivePenguin !

Issue #68, opened 2023-02-20
Represent the supported languages as an enum


Czocher:
Currently some parts of the code, like [here](https://github.com/1Password/typeshare/blob/main/core/src/rust_types.rs#L41), represent the supported languages as a `String`. In reality this string will always be one of the supported language names - `typescript`, `swift`, `go`, `kotlin`.
We should represent the supported languages as an `enum` instead.

InquisitivePenguin:
Fixed in #72

PR #72, opened 2023-03-06
Changed supported languages from `String` to `enum`

DuhPesky:
Potential fix for #68 

Instead of using `String` for the supported language names (e.g. `scala` or `kotlin`), we now represent the supported languages as ```enum SupportedLanguage```.  

Didn't change ALL of the `cli` crate code to use it, because at a minimum we would need to impl some clap traits or convert to &str in const context. Which means adding clap as a dependency to `core` or doing `impl const From<SupportedLanguage> for &str`, but const impl's are experimental. If there are other ways, I can implement it.

### DIFF
--- a/core/src/language/mod.rs
+++ b/core/src/language/mod.rs
@@ -1,9 +1,10 @@
 use crate::{
-    parser::ParsedData,
+    parser::{ParseError, ParsedData},
     rust_types::{Id, RustEnum, RustEnumVariant, RustStruct, RustTypeAlias},
 };
 use itertools::Itertools;
-use std::{collections::HashMap, io::Write};
+use proc_macro2::Ident;
+use std::{collections::HashMap, fmt::Debug, io::Write, str::FromStr};
 
 mod go;
 mod kotlin;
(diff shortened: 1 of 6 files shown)


core/src/language/swift.rs
```diff
@@ -210,9 +210,9 @@ impl Language for Swift {
         let mut decs = self.get_default_decorators();
 
         // Check if this struct's decorators contains swift in the hashmap
-        if rs.decorators.contains_key("swift") {
+        if rs.decorators.contains_key(&SupportedLanguage::Swift) {
             // this unwrap is safe since we just checked if we contain the key
-            let swift_decs = rs.decorators.get("swift").unwrap();
+            let swift_decs = rs.decorators.get(&SupportedLanguage::Swift).unwrap();
```
CerulanLumina, line 215: This could be improved by using `if let Some(...)`

CerulanLumina (changes requested):
Looks good to me so far, just requesting a minor change to improve our code quality. Thanks for your submission!

core/src/language/mod.rs
```diff
@@ -18,6 +19,39 @@ pub use scala::Scala;
...
 
+/// All supported languages.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum SupportedLanguage {
+    Go,
+    Kotlin,
+    Scala,
+    Swift,
+    TypeScript,
+}
```
CerulanLumina, line 30: We use `#[deny(missing_docs)]`, so these variants need documentation. They're self-explanatory though, so I'd have no problem using `#[allow(missing_docs)]` on this enum.

core/src/language/swift.rs
DuhPesky: done.

core/src/language/mod.rs
DuhPesky: Yeah, there isn't really too much to document here, so just went with your allow lint. done!

CerulanLumina (approved):
Looks good to me, thanks for your contributions!

PR #75, opened 2023-03-12
Add Scala support to CLI

jclmnop:
# Motivation
When reading the docs for `typeshare`, I noticed that Scala was mentioned in the supported languages but it wasn't mentioned in the docs for the cli. 

Dug into the source code a bit and could see Scala was missing from `cli/src/main.rs`.

I figure there are two main possibilities:

1. Support for Scala isn't actually complete so you don't want to include it in the cli yet
2. After implementing Scala in the core crate, you forgot to add it to the cli crate. You guys are busy, after all. 

If it's the first option, then it's not a big issue because writing this PR didn't take very long anyway and you can just close it. 

If it _is_ the second option and this PR helps, then I'd propose a refactor/rewrite of the CLI (although I think it needs one anyway). In its current state I'd expect it's tricky to keep track of. 

I'd go for a more `struct`ured approach (😏), if you look at the sorts of features `clap` has integrated from [`structopt`](https://docs.rs/structopt/latest/structopt/) you'll see what I mean. Happy to open up a RFC issue with a more detailed proposal if you're interested, but I'm also happy to just crack on with it next weekend without an RFC if that's alright with you guys. I'd also look into adding some proper tests for the actual CLI functionality. 

# Changes
5734f27e6defc3f866a35413b75536a5fe810206: Added `"scala"` as a supported value for the `--lang` arg, along with args and overrides for `config.scala.package` and `config.scala.module_name`

5a7de2f93830b00ae8772ab967ee2b91e77faaaf: Added Scala values to the test configs

04c1a478fffc36d0d247ab693c56349cee62680c: Added mentions of Scala in the cli README and also in the book. Quite possible I've missed out some places where it should be mentioned though.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -88,7 +88,6 @@ fn build_command() -> Command<'static> {
         )
         .arg(
             Arg::new(ARG_SCALA_PACKAGE)
-                .short('S')
                 .long("scala-package")
                 .help("Scala package name")
                 .takes_value(true)
@@ -96,7 +95,6 @@ fn build_command() -> Command<'static> {
         )
         .arg(
             Arg::new(ARG_SCALA_MODULE_NAME)
-                .short('M')
                 .long("scala-module-name")
                 .help("Scala serializer module name")
                 .takes_value(true)
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -14,6 +14,7 @@ use typeshare_core::{
     language::{Kotlin, Language, Swift, TypeScript},
     parser::ParsedData,
 };
+use typeshare_core::language::Scala;
 
 mod config;
 
@@ -23,6 +24,8 @@ const ARG_TYPE: &str = "TYPE";
 const ARG_SWIFT_PREFIX: &str = "SWIFTPREFIX";
 const ARG_JAVA_PACKAGE: &str = "JAVAPACKAGE";
 const ARG_MODULE_NAME: &str = "MODULENAME";
+const ARG_SCALA_PACKAGE: &str = "SCALAPACKAGE";
+const ARG_SCALA_MODULE_NAME: &str = "SCALAMODULENAME";
 #[cfg(feature = "go")]
 const ARG_GO_PACKAGE: &str = "GOPACKAGE";
 const ARG_CONFIG_FILE_NAME: &str = "CONFIGFILENAME";
@@ -30,10 +33,10 @@ const ARG_GENERATE_CONFIG: &str = "generate-config-file";
 const ARG_OUTPUT_FILE: &str = "output-file";
 
 #[cfg(feature = "go")]
-const AVAILABLE_LANGUAGES: [&str; 4] = ["kotlin", "swift", "typescript", "go"];
+const AVAILABLE_LANGUAGES: [&str; 5] = ["kotlin", "scala", "swift", "typescript", "go"];
 
 #[cfg(not(feature = "go"))]
-const AVAILABLE_LANGUAGES: [&str; 3] = ["kotlin", "swift", "typescript"];
+const AVAILABLE_LANGUAGES: [&str; 4] = ["kotlin", "scala", "swift", "typescript"];
 
 fn build_command() -> Command<'static> {
     command!("typeshare")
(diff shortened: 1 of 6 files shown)


InquisitivePenguin:
We didn't enable support for Scala immediately because there were some incomplete elements - at least that was the impression I got from the contributor who wrote the original PR. But there have been some follow-up PRs, so we should be good to enable Scala support now. I'll review your code when I get the chance.

@exoego you should probably take a look at this too.

cli/src/main.rs
```diff
@@ -14,6 +14,7 @@ use typeshare_core::{
     language::{Kotlin, Language, Swift, TypeScript},
     parser::ParsedData,
 };
+use typeshare_core::language::Scala;
```
exoego, line 17: Not sorted.

jclmnop: Whoops. Fixed and rebased.

exoego (approved):
LGTM
Though I am just an initial contributor for Scala support

jclmnop:
> We didn't enable support for Scala immediately because there were some incomplete elements - at least that was the impression I got from the contributor who wrote the original PR. But there have been some follow-up PRs, so we should be good to enable Scala support now. I'll review your code when I get the chance.

Yeah that's fair enough.

Also I forgot to mention this in the PR, but one thing I wasn't sure of, is whether new args were actually needed for the `config.scala.package` and `config.scala.module_name` overrides. I added them, but if it makes more sense to just use the existing args for Kotlin (with an update description mentioning that they're used for Scala also) then I can quickly make that change. 

My reasoning for adding new args, is in case someone is generating a config and they want different values for these fields in their Scala and Kotlin configs, or if someone is running multiple commands to generate types for a few languages but they just want to specify the overrides once and change the `--lang` arg each time they use the command.

InquisitivePenguin:
@jclmnop I merged `main` into your branch to fix some conflicts. Reviewing now.

docs/src/usage/usage.md
```diff
@@ -9,6 +9,7 @@ To generate ffi definitions for a specific target language, run the `typeshare`
 typeshare ./my_rust_project --lang=kotlin --output-file=my_kotlin_definitions.kt
 typeshare ./my_rust_project --lang=swift --output-file=my_swift_definitions.swift
 typeshare ./my_rust_project --lang=typescript --output-file=my_typescript_definitions.ts
+typeshare ./my_rust_project --lang=sca;a --output-file=my_scala_definitions.scala
```
InquisitivePenguin, line 12: Looks like you made a typo here - should be `scala` instead of `sca;a`

cli/src/main.rs
```diff
@@ -84,6 +86,22 @@ fn build_command() -> Command<'static> {
                 .takes_value(true)
                 .required(false),
         )
+        .arg(
+            Arg::new(ARG_SCALA_PACKAGE)
+                .short('S')
```
InquisitivePenguin, line 91: Honestly I think we could go without the short args here, especially since it's really unclear what this is an abbreviation for at a glance. Also, the lowercase version is for Swift prefixes, which is totally unrelated and having the capitalized version be a config for a different language _and_ option is confusing.

cli/src/main.rs
```diff
@@ -84,6 +86,22 @@ fn build_command() -> Command<'static> {
...
+                .long("scala-package")
+                .help("Scala package name")
+                .takes_value(true)
+                .required(false),
+        )
+        .arg(
+            Arg::new(ARG_SCALA_MODULE_NAME)
+                .short('M')
```
InquisitivePenguin, line 99: Same issue here, there's no clear reason why uppercase would mean 'Scala module name' and lowercase would mean 'Kotlin module name'. I propose that we get rid of both short args, although we should probably hold off on changing the Kotlin args right now since that would be a breaking change.

InquisitivePenguin (changes requested):
Some minor nitpicks but otherwise it's looking good!

InquisitivePenguin:
> Also I forgot to mention this in the PR, but one thing I wasn't sure of, is whether new args were actually needed for the config.scala.package and config.scala.module_name overrides. I added them, but if it makes more sense to just use the existing args for Kotlin (with an update description mentioning that they're used for Scala also) then I can quickly make that change.

Ideally we should keep those separate - the package/module names could be completely different between the two languages.

cli/src/main.rs
jclmnop: I agree, I just couldn't think of anything clearer and wanted to follow the previous convention of including short args.

jclmnop:
> Some minor nitpicks but otherwise it's looking good!

Fixed 👍

InquisitivePenguin (approved):
Thank you for getting this done!

PR #78, opened 2023-03-13
Attempt to fix generated Swift code for deserializing optional associated types in enum variants

### LINKED ISSUE #77: Deserialization fails in Swift for enums containing optional values
Suppose we have an enum with a variant containing an optional value:
```rust
#[typeshare]
#[derive(Serialize, Deserialize)]
pub enum AlgebraicEnumWithOptionalValue {
  Variant(Option<()>)
}
```

The code to deserialize this variant will look like this:
```
			case .variant:
				if let content = try? container.decode(AlgebraicEnumWithOptionalValue?.self, forKey: .content) {
					self = .variant(content)
					return
				}
```

However, if `content` is null (i.e. the variant on the Rust side contained `None`), this will fail to deserialize with error 1002.

InquisitivePenguin:
Fixes #77, see the issue for details.

### DIFF
--- a/core/src/language/swift.rs
+++ b/core/src/language/swift.rs
@@ -520,6 +520,7 @@ impl Swift {
                             ));
                         }
                         RustEnumVariant::Tuple { ty, .. } => {
+                            let content_optional = ty.is_optional();
                             let case_type = self
                                 .format_type(ty, e.shared().generic_types.as_slice())
                                 .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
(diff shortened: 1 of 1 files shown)


core/src/language/swift.rs
```diff
@@ -529,12 +530,17 @@ impl Swift {
                                 "
 			case .{case_name}:
 				if let content = try? container.decode({case_type}.self, forKey: .{content_key}) {{
```
kevinbhayes, line 532: I believe this will fail if content is null. I think we need an else if case to capture that.

```
			case .{case_name}:
				if let content = try? container.decode({case_types}.self, forKey: .content) {
					self = .{case_name}(content)
					return
				}
				else if let isNil = try? container.decodeNil(forKey: .content), isNil {
					self = .{case_name}(nil)
					return
				} 

```

kevinbhayes (changes requested):

core/src/language/swift.rs
InquisitivePenguin: Should be fixed in the latest commit.

CerulanLumina (approved):

Issue #64, opened 2023-02-13
Generate pre-built binaries


oeb25:
Hey! I've been using `typeshare` happily, but it's become a pain point having to build `typeshare-cli` from source in CI.

I use [`taiki-e/install-action@v2`](https://github.com/marketplace/actions/install-development-tools) and [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall) which both allow downloading binaries from different archives.

To generate these binaries I can recommend [`cargo-dist`](https://github.com/axodotdev/cargo-dist), which builds binaries for multiple platforms, and stores them in a GitHub Release where tools like `cargo-binstall` knows where to find them.

InquisitivePenguin:
Thanks @oeb25! We will definitely look into doing this.

InquisitivePenguin:
We now have `cargo-dist` configured for this repository. @oeb25 Could you try out the prereleases that have been published and see if they work for you?

InquisitivePenguin:
I'm going to mark this as done. @oeb25 Feel free to re-open this issue if it's not working as intended for you

oeb25:
Hey sorry I've been away!

First of all, thanks for doing this!

I tried doing the install, but it unfortunately fails. I think it might be due to the binary crate being named differently to the binary and generated artifact or something. Here's the command:

```sh
# With cargo-binstall v0.19.3
❯ cargo binstall typeshare-cli
 INFO resolve: Resolving package: 'typeshare-cli'
 WARN resolve: Error while checking fetcher invalid url: could not GET https://github.com/1Password/typeshare/releases/download/v1.1.0/typeshare-cli_aarch64-apple-darwin_1.1.0.tar.xz: error sending request for url (https://github.com/1Password/typeshare/releases/download/v1.1.0/typeshare-cli_aarch64-apple-darwin_1.1.0.tar.xz): error trying to connect: dns error: proto error: io error: Too many open files (os error 24)
 WARN resolve: Error while checking fetcher invalid url: could not GET https://github.com/1Password/typeshare/releases/download/v1.1.0/typeshare-cli_x86_64-apple-darwin_1.1.0.tzstd: error sending request for url (https://github.com/1Password/typeshare/releases/download/v1.1.0/typeshare-cli_x86_64-apple-darwin_1.1.0.tzstd): error trying to connect: dns error: proto error: io error: Too many open files (os error 24)
[...]

PR #79, opened 2023-03-14
Update release config with cargo-dist@v0.0.4

oeb25:
This is generated using cargo-dist@v0.0.4 and running:

```sh
cargo dist init --ci=github --installer=shell --installer=powershell
```

This change introduces a new target, namely `aarch64-apple-darwin`, and also generates installer scripts for shell and powershell. Both of these additions can be turned off if needed.

Hopefully, this change might fix some of the issues with cargo-binstall not finding the binary. (See #64 for context)

The toolchain used for building in release is also fixed to 1.57.1, same as it was in the prior version of the release CI.

I didn't notice any custom configuration made to the files that should be ported over, but I might have missed some.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,24 @@ members = [
     "lib"
 ]
 
+# Config for 'cargo dist'
+[workspace.metadata.dist]
+# The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
+cargo-dist-version = "0.0.4"
+# The preferred Rust toolchain to use in CI (rustup toolchain syntax)
+rust-toolchain-version = "1.57.1"
+# CI backends to support (see 'cargo dist generate-ci')
+ci = ["github"]
+# The installers to generate for each app
+installers = ["shell", "powershell"]
+# Target platforms to build apps for (Rust target-triple syntax)
+targets = [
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -16,14 +34,7 @@ debug-assertions = false
...
-[workspace.metadata.dist]
-# The preferred Rust toolchain to use in CI (rustup toolchain syntax)
-# Here, we use the MSRV.
-rust-toolchain-version = "1.57.1"
-# CI backends to support (see 'cargo dist generate-ci')
-ci = ["github"]
-# Target platforms to build apps for (Rust target-triple syntax)
-targets = ["x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows-msvc"]
\ No newline at end of file
+lto = "thin"
```
InquisitivePenguin, line 40: I'm not against changing our LTO config, but I think we should move this to our default `profile.release` for consistency.

InquisitivePenguin (approved):
LGTM, I just have one question.

Cargo.toml
oeb25: That would make sense yes! This was just the default generated by `cargo dist init`. Do you want me to move it to `profile.release` in this PR?

PR #81, opened 2023-03-21
Remove version headers from snapshot tests

### LINKED ISSUE #73: Updating the version of `typeshare-cli` breaks all existing tests
Running `UPDATE_EXPECT=1 cargo test` every time the version gets updated is time-consuming, breaks PRs that introduce new tests, and can introduce false positives if done incorrectly. Our diff test system should be able to account for the version comment header changing, and ideally the snapshot tests shouldn't have it at all.

InquisitivePenguin:
This will no longer cause test breakage when the version is updated.

A consequence of this PR is the introduction of a new public variable for all current language implementation constructors.

Fixes #73.


CerulanLumina (approved):
Looks good.

2Pacalypse-:
Hey @snowsignal, sorry for the necro, but would it be possible to expose this `no_version_header` option so we can use it in the `typeshare.toml` configuration file? Thank you!

PR #82, opened 2023-03-23
Generated Typescript now uses `null` to represent unit types

### LINKED ISSUE #43: Typescript: `unit` seems to be wrongly defined as `undefined` in Typescript
Hi,

I've the following Rust code:

```rust
#[typeshare::typeshare]
pub enum Cmd {
    StartBrowser(StartBrowser),
}

#[typeshare::typeshare]
pub struct StartBrowser(pub ());
```

...which generates the following Typescript code:

```typescript
export type StartBrowser = undefined;

export type Cmd = 
	| { type: "StartBrowser", content: StartBrowser };
```

...however, `serde_json` serializes `Cmd::StartBrowser(StartBrowser(()))` as `{"type":"StartBrowser","content":null}` (please note the `null` instead of the above defined `undefined`).

I'm not 100% sure, but I believe `serde_json` uses `undefined` to represent optional values (i.e., Rust's `Option` type).

InquisitivePenguin:
`()` is represented as `null` instead of `undefined` - this also applies to unit enum variants.

Fixes #43.

### DIFF
--- a/core/data/tests/boxed_value/output.ts
+++ b/core/data/tests/boxed_value/output.ts
@@ -1,6 +1,6 @@
 /** This is a comment. */
 export type Colors = 
-	| { type: "Red", content?: undefined }
-	| { type: "Blue", content?: undefined }
+	| { type: "Red", content: null }
+	| { type: "Blue", content: null }
 	| { type: "Green", content: string };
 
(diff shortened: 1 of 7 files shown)


lazgilebits (approved):
Confirming all of the newly generated code is valid TypeScript 👍 

Approved!

CerulanLumina (approved):

PR #85, opened 2023-03-24
the cooler Date types PR

peter-lyons-kehl:
@CerulanLumina This is your `support-date-types` rebased on `main`, with minor TypeScript and Kotlin extras. Use as you like.

### DIFF
--- a/core/src/language/typescript.rs
+++ b/core/src/language/typescript.rs
@@ -5,6 +5,7 @@ use crate::{
     language::{Language, SupportedLanguage},
     rust_types::{RustEnum, RustEnumVariant, RustField, RustStruct, RustTypeAlias},
 };
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{collections::HashMap, io::Write};
 
 /// All information needed to generate Typescript type-code
@@ -15,6 +16,7 @@ pub struct TypeScript {
     /// Whether or not to exclude the version header that normally appears at the top of generated code.
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
+    pub has_date: AtomicBool,
 }
 
 impl Language for TypeScript {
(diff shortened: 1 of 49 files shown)


peter-kehl:
Not ready for merge, but I'm putting this out Draft, so you get CI run the tests.

CerulanLumina:
Hey, thanks for this! I had gotten sidetracked on other projects so I haven't looked at this branch in a while.

core/src/language/typescript.rs
```diff
@@ -84,6 +92,15 @@ impl Language for TypeScript {
         Ok(())
     }
 
+    fn end_file(&mut self, w: &mut dyn Write) -> std::io::Result<()> {
+        if self.has_date.load(Ordering::SeqCst) {
+            writeln!(w, "export function TypeshareDateReviver(key, value): Date {{ return new Date(value); }}")?;
```
InquisitivePenguin, line 97: What is the purpose for this function? Why does it need to take `key`?

core/src/language/typescript.rs
```diff
@@ -15,6 +16,9 @@ pub struct TypeScript {
     /// Whether or not to exclude the version header that normally appears at the top of generated code.
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
+    /// Whether the generated code includes a date type
+    /// (this means we need to output a reviver function as well)
+    pub has_date: AtomicBool,
```
InquisitivePenguin, line 21: Now that `Language` handles codegen with a mutable reference, this can probably just be a simple `bool`.

InquisitivePenguin (changes requested):

core/src/language/typescript.rs
CerulanLumina: This is a provided function that can be passed to `JSON.parse` to revive JSON into an actual `Date` type rather than just an object. The function signature is necessary as it is what `JSON.parse` expects

core/src/language/typescript.rs
CerulanLumina: Fixed :)

Exotik850:
Is there any work left to be done for this? Would be quite useful for many applications

rt-planck:
Awesome that there is an ongoing fix.
What is missing here? How can we help to accelerate merging this bugfix?

hculea:
@rt-planck @Exotik850 @peter-lyons-kehl I wanted to signal-boost [this pull request](https://github.com/1Password/typeshare/pull/236) which is achieves the same end for Go, Typescript and Python. 

We are more than welcoming contributions for the other languages 😄 

This should be available with the next Typeshare release. - given that this has been in draft for two years and !236 achieves the same end, I will be closing this PR. Happy to hear your feedback on the dates work!

Issue #32, opened 2022-12-14
Unexpected token when parsing type: `[u16 ; 9]`


takkuumi:
thread '<unnamed>' panicked at 'failed to parse a rust type: Unexpected token when parsing type: `[u16 ; 9]`. This is an internal error, please ping a typeshare developer to resolve this problem.', /Users/takumi/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/typeshare-cli-1.0.1/src/main.rs:234:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

CerulanLumina:
Hello! Typeshare currently does not support fixed size arrays. For now, use `Vec<u16>` instead. The returned panic message is misleading though - I'll create a new issue to improve this.

NateLing:
@CerulanLumina could it map this to `number[]`

InquisitivePenguin:
Closed by #89

PR #89, opened 2023-04-06
Add support for fixed length arrays

ccouzens:
These have been a problem in a couple structures I'm hoping to translate into Typescript.

I'm unsure of the syntax for all the languages excepting typescript. For the majority of them I reused Vec (because a fixed length array is a specialised version of a dynamically sized array).

For go, I found something that suggested this syntax `[4]string`. I don't know if this is right, so I'd appreciate input on the PR.

I would appreciate input in the PR for go, kotlin, scala and swift.

### DIFF
--- a/core/data/tests/generate_types/output.ts
+++ b/core/data/tests/generate_types/output.ts
@@ -8,6 +8,7 @@ export interface Types {
 	float: number;
 	double: number;
 	array: string[];
+	fixed_length_array: [string, string, string, string];
 	dictionary: Record<string, number>;
 	optional_dictionary?: Record<string, number>;
 	custom_type: CustomType;
--- a/core/data/tests/generate_types/input.rs
+++ b/core/data/tests/generate_types/input.rs
@@ -9,6 +9,7 @@ pub struct Types {
     pub float: f32,
     pub double: f64,
     pub array: Vec<String>,
+    pub fixed_length_array: [String; 4],
     pub dictionary: HashMap<String, i32>,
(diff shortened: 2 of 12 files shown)


ccouzens:
This would resolve https://github.com/1Password/typeshare/issues/32

core/data/tests/generate_types/output.ts
```diff
@@ -8,6 +8,7 @@ export interface Types {
 	float: number;
 	double: number;
 	array: string[];
+	fixed_length_array: [string, string, string, string];
```
InquisitivePenguin, line 11: This is idiomatic, though I do worry that large values of `n` could blow up the file size significantly. I wonder if we could conditionally define an n-length tuple like so:

```ts
type FixedArray<T, N extends number> = N extends N ? number extends N ? T[] : _FixedArrayOf<T, N, []> : never;
type _FixedArrayOf<T, N extends number, R extends unknown[]> = R['length'] extends N ? R : _FixedArrayOf<T, N, [T, ...R]>;
```
and then use `Tuple<string, 4>`.

(taken from [this Stack Overflow answer](https://stackoverflow.com/a/52490977))

I can work on this in a follow-up.

InquisitivePenguin:
Thank you @ccouzens for this MR!

`[4]string` is indeed correct Go syntax. I ran the updated examples through the Go playground to confirm this.

I have a small suggestion with Typescript codegen but that can be fixed in a follow-up.

InquisitivePenguin (approved):

PR #95, opened 2023-04-11
Configurable Generic Constraints for Swift

### LINKED ISSUE #86: Support specifying swift Generic Type Constraints
Consider the following struct:

```
#[typeshare]
#[derive(Serialize, Debug)]
#[serde(rename_all = "camelCase")]
pub struct AccountPayload<A> {
    pub id: String,
    pub payload: A
}
```

It is generated as: (we have Sendable in the typeshare.toml file)
```
public struct AccountPayload<A: Codable>: Codable, Sendable {
	public let id: String
	public let payload: A
…
	}
}
```

However, generic type `A` (and others) may also have another type constraint (in this case `Sendable` supplied) so that it needs to be generated as:
```
public struct CoreAccountMenuEntry<A: Codable & Sendable>: Codable, Sendable {
…
```

### Suggestion:

An additional decorator parameter in the toml:
`default_generic_decorators = ["Sendable"]`

And/or in the rust definitions itself:
`[typeshare(swift="Whatever", swift_generic="A:Sendable & AnotherProtocol,B:Sendable & SomeOtherProtocol")]` where A and B are generics specified in the struct. (I added B here in case there is more than one generic)

InquisitivePenguin:
Resolves #86.

Default type constraints can now be specified for Swift within the config, using `default_generic_constraints = [...]` under `[swift]`. Constraint strings within the list can include multiple constraints, such as `Sendable & Identifiable`.

### DIFF
--- a/cli/data/tests/constraints_config.toml
+++ b/cli/data/tests/constraints_config.toml
@@ -1,0 +1,2 @@
+[swift]
+default_generic_constraints = ["Sendable"]
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -31,17 +31,17 @@ pub struct SwiftParams {
     pub prefix: String,
     pub type_mappings: HashMap<String, String>,
     pub default_decorators: Vec<String>,
-    pub default_generic_decorators: Vec<String>,
+    pub default_generic_constraints: Vec<String>,
 }
 
 impl Default for SwiftParams {
     fn default() -> Self {
-        let default_generic_decorators = vec!["Codable".into()];
+        let default_generic_constraints = vec!["Codable".into()];
         Self {
             prefix: Default::default(),
             type_mappings: Default::default(),
             default_decorators: Default::default(),
-            default_generic_decorators,
+            default_generic_constraints,
         }
     }
 }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -25,12 +25,25 @@ pub struct ScalaParams {
     pub type_mappings: HashMap<String, String>,
 }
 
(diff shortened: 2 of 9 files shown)


CerulanLumina (commented):
Looks good so far, needs further tests as mentioned.

kevinbhayes (commented):
The default typeshare.toml specification worked, but if I didn't specify it there, and tried to apply it to a single type, it did nothing. For example:

```
#[typeshare(swift(generic_decorators = "Sendable & Codable"))]
#[derive(Serialize, Debug)]
#[serde(rename_all = "camelCase")]
pub struct MyMenuEntry<A> {
```
generated:
```
public struct MyMenuEntry<A: Codable>: Codable, Sendable
```

InquisitivePenguin:
@kevinbhayes I've been trying to fix this - it looks like parenthesized parameters like `swift(...)` were being ignored by the Swift codegen code, and this issue runs deeper than I expected.

kevinbhayes:
> @kevinbhayes I've been trying to fix this - it looks like parenthesized parameters like `swift(...)` were being ignored by the Swift codegen code, and this issue runs deeper than I expected.

If you want to break that part off into a separate issue, that's ok with me. I think we'll be using the toml definitions anyway.

InquisitivePenguin:
@kevinbhayes Okay, I'll split this PR.

kevinbhayes (approved):
Tested by adding Sendable to `default_generic_constraints = ["Sendable"]` and it produced `MyStruct<A: Codable & Sendable>: Codable` so all is good!

Issue #96, opened 2023-04-12
null content is no longer optional


chrishoage:
https://github.com/1Password/typeshare/pull/82/

This PR changed `undefined` to `null` however it removed the optional operator (from `content?: undefined` to `content: null`)

This seems like a change that should at the very least be a setting. 

This is useful if a rust FFI interface allows optional parameters to be passed. 

Before one could express an interface that did not require `content: null` to be passed (as this is useless) and now it is required. 

I propose `content?: null` be allowed. While this does mean that the type is effectively `content?: null | undefined` this allows for a better quality of life when allowing to not send an empty content over FFI

PR #98, opened 2023-04-13
Draft: Change the representation of unit types

CerulanLumina:
This was previously fixed though its merge highlighted several issues that have yet to be resolved. See discussion at the following locations.

#43 
#82 
#96 

This is the new PR to fix this which helpfully already includes the necessary 'revert revert' commit.


Lucretiel:
Per #96, it seems like the relevant types should be `name?: null` instead of `name: null`; was that reverted deliberately?

PR #99, opened 2023-04-14
Set up a nix flake that links all of the typeshare releases

savannidgerinel:
This allows Nix users to include Typeshare in their projects directly, or even to run typeshare on the command line. We're also exporting separate packages for each version of typeshare, so that users can choose the one that suits them best, or even pin to an exact version as a hedge against changes.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,60 @@
+{
+  "nodes": {
+    "flake-utils": {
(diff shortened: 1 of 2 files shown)


savannidgerinel:
It has been prevailed upon me that this is the wrong strategy for long-term life of a flake.

PR #104, opened 2023-04-17
Set up a nix flake for typeshare

savannidgerinel:
This allows Nix users to include Typeshare in their projects directly, or even to run typeshare on the command line. Nix users will be able to pin against past releases of the package by setting a version number or even a particular commit in their `flake.nix` file when importing the Typeshare flake.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,60 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1681202837,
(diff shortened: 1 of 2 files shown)


CerulanLumina (changes requested):
All looks good as far as I can tell (I don't use NixOS personally). Only one thing: your commit seems to have failed signing verification. Once that is fixed, I'll merge this!

savannidgerinel:
@CerulanLumina Uploaded my GPG key and my commit is now verified. All yours!

CerulanLumina (approved):

PR #109, opened 2023-04-25
Typeshare can now detect types within embedded modules

### LINKED ISSUE #40: Typeshare ignores type annotations in embedded modules
Hi,

I have code like this:

```rust
mod my_module {
    // Types with Typeshare annotations
}
```

...which results in nothing being generated.

If I remove the module around the annotations, everything works.

InquisitivePenguin:
Say, for instance, you have this situation:

```
mod inline_module {
  #[typeshare]
  pub struct Inline;
}
```

Previously, `Inline` would not be detected since our parser was not scanning the contents of inline/embedded modules. This has been fixed.

Resolves #40

### DIFF
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -110,7 +110,7 @@ pub fn parse(input: &str) -> Result<ParsedData, ParseError> {
 fn flatten_items<'a>(items: impl Iterator<Item=&'a syn::Item>) -> impl Iterator<Item=&'a syn::Item> {
     items.flat_map(|item| match item {
         syn::Item::Mod(syn::ItemMod { content: Some((_, items)), .. }) => {
-            items.iter().collect()
+            flatten_items(items.iter()).collect()
         },
         item => vec![item]
     }.into_iter())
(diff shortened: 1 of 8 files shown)


core/src/parser.rs
```diff
@@ -106,6 +107,23 @@ pub fn parse(input: &str) -> Result<ParsedData, ParseError> {
...
+fn flatten_items<'a>(
+    items: impl Iterator<Item = &'a syn::Item>,
+) -> impl Iterator<Item = &'a syn::Item> {
+    items.flat_map(|item| {
+        match item {
+            syn::Item::Mod(syn::ItemMod {
+                content: Some((_, items)),
+                ..
+            }) => flatten_items(items.iter()).collect(),
+            item => vec![item],
```
CerulanLumina, line 121: Wish we could avoid allocs here but I don't think we can without doing something incredibly cursed

CerulanLumina (approved):

core/src/parser.rs
InquisitivePenguin: Unfortunately ☹️

Issue #24, opened 2022-11-27
Are `usize` and `i64` not supported?


affanshahid:
Ran typeshare on my project and got the following errors:

```
thread '<unnamed>' panicked at 'failed to parse a rust type: ["usize"]', /Users/affan/.cargo/registry/src/github.com-1ecc6299db9ec823/typeshare-cli-1.0.0/src/main.rs:209:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'failed to parse a rust type: ["i64"]', /Users/affan/.cargo/registry/src/github.com-1ecc6299db9ec823/typeshare-cli-1.0.0/src/main.rs:209:17
```

i64 can be represented using a JS BigInt.

jeremychone:
as with `ts-rs` would be nice to have a `#[typeshare(type = "number")]` property attribute.

InquisitivePenguin:
Thanks for opening this issue. I'll look into whether this is possible.

BenJeau:
Getting a similar error with `u64` and `isize`, those four types seems to be "disabled" here

https://github.com/1Password/typeshare/blob/fbf5aea145b7895f3a998db050bd972a9af79e05/core/src/rust_types.rs#L225-L227

But the types exist in the [SpecialRustType](https://github.com/1Password/typeshare/blob/fbf5aea145b7895f3a998db050bd972a9af79e05/core/src/rust_types.rs#L108) and looking in the four supported languages, they all have mappings for those four types.

Locally I've changed those three lines in `rust_types.rs` to parse the types and I'm not getting any errors, but I'm guessing that there must be a reason that those lines/errors exist? It works for go, swift, typescript, and kotlin.

### Changes

In `rust_types.rs`:

```diff
-                    "u64" | "i64" | "usize" | "isize" => {
-                        return Err(RustTypeParseError::UnsupportedType(vec![id]))
-                    }
+                    "u64" => RustType::Special(SpecialRustType::U64),
+                    "usize" => RustType::Special(SpecialRustType::USize),
+                    "i64" => RustType::Special(SpecialRustType::I64),
+                    "isize" => RustType::Special(SpecialRustType::ISize),
```

and in `typescript.rs`:

```diff
            SpecialRustType::U64
            | SpecialRustType::I64
            | SpecialRustType::ISize
            | SpecialRustType::USize => {
-                panic!("64 bit types not allowed in Typeshare")
+                Ok("number".into())
            }
```

[...]

PR #112, opened 2023-04-30
Add support for typescript bit-int types

annie444:


annie444:
NOTE: I can't seem to find a logger or default stream setup for writing to stdin, so I just opted for println!().

annie444:
This would close 

#24

core/src/language/typescript.rs
```diff
@@ -69,7 +71,21 @@ impl Language for TypeScript {
             | SpecialRustType::I64
             | SpecialRustType::ISize
             | SpecialRustType::USize => {
-                panic!("64 bit types not allowed in Typeshare")
+                if self.unsafe_types {
+                    println!(
+                        r#"WARNING: 
+        64bit types are not supported by the default JSON lib for JS/TS! This will lead to memory safety issues when passing data between JS/TS and Rust.
+        If your intended encpoint is for JSON parsing/serializing between JS/TS & Rust, please consider using a custom JSON parser that supports 64 bit static types.
```
czocher, line 78: Typo - encpoint should be endpoint I believe?

eboody:
I think testing is going to continue to fail until you remove the big-int-ish types from the list of blocklisted types:

https://github.com/1Password/typeshare/blob/main/core/src/rust_types.rs#L257

Issue #114, opened 2023-05-06
Suggestion for #[serde(flatten)]


Roms1383:
Hi, just tried your library for the first time and really like it, also love its simplicity.

What lead me here is that I tried `#[serde(flatten)]` and got the error.
Here's one suggestion that would work pretty well IMHO, for [serde struct flattening](https://serde.rs/attr-flatten.html),
given for example this code:

```rs
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[typeshare]
pub struct Geolocation {
    pub lat: f32,
    pub long: f32,
}

#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[typeshare]
pub struct Meta {
    #[serde(rename = "createdAt")]
    pub created_at: chrono::DateTime<chrono::Utc>,
    #[serde(rename = "updatedAt")]
    pub updated_at: chrono::DateTime<chrono::Utc>,
}

#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[typeshare]
pub struct User {
    email: String,
    #[serde(flatten)]
    geo: Geolocation,
    #[serde(flatten)]
    meta: Meta,
}
```
would generate:
```ts
export interface Geolocation {
	lat: number;
	long: number;
}

export interface Meta {
	createdAt: Date;
	updatedAt: Date;
}

export interface User extends Geolocation, Meta {
	email: string;
}
```
I thought that would be a pretty good fit, especially since in Typescript an interface can inherit multiple interfaces, and it also does flatten the nested fields in the parent definition.
Let me know what you think, thanks for reading me :)

PR #115, opened 2023-05-07
Support serde flatten for Typescript

Roms1383:
This PR concretizes #114.
I haven't added tests, yet.

### DIFF
--- a/core/src/language/typescript.rs
+++ b/core/src/language/typescript.rs
@@ -160,6 +160,10 @@ impl Language for TypeScript {
             }
         }
     }
+
+    fn supports_flatten(&self) -> bool {
+        true
+    }
 }
 
 impl TypeScript {
--- a/core/src/language/typescript.rs
+++ b/core/src/language/typescript.rs
@@ -110,13 +110,33 @@ impl Language for TypeScript {
 
     fn write_struct(&mut self, w: &mut dyn Write, rs: &RustStruct) -> std::io::Result<()> {
         self.write_comments(w, 0, &rs.comments)?;
+        let mut inheritance = "".to_string();
+        let mut count = 0;
+        for field in rs.fields.iter() {
+            if field.flattened {
+                let ts_ty = match self.format_type(&field.ty, rs.generic_types.as_slice()) {
+                    Ok(v) => v,
+                    Err(e) => return Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, e)),
+                };
+                if count >= 1 {
+                    inheritance.push_str(", ");
+                }
+                inheritance.push_str(ts_ty.as_str());
+                count += 1;
+            }
+        }
         writeln!(
             w,
-            "export interface {}{} {{",
+            "export interface {}{}{} {{",
             rs.id.renamed,
             (!rs.generic_types.is_empty())
                 .then(|| format!("<{}>", rs.generic_types.join(", ")))
-                .unwrap_or_default()
+                .unwrap_or_default(),
+            if !inheritance.is_empty() {
+                format!(" extends {inheritance}")
+            } else {
+                "".to_string()
+            }
         )?;
 
         rs.fields
@@ -234,6 +254,9 @@ impl TypeScript {
         field: &RustField,
         generic_types: &[String],
     ) -> std::io::Result<()> {
+        if field.flattened {
+            return Ok(());
+        }
         self.write_comments(w, 1, &field.comments)?;
         let ts_ty = self
             .format_type(&field.ty, generic_types)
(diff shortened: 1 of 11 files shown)


Roms1383:
So, before I figure out how tests are setup,
here's what it does:

```rs
#[derive(Debug, Deserialize, Serialize)]
#[typeshare]
pub struct Foo {
    pub value: String,
}

#[derive(Debug, Deserialize, Serialize)]
#[typeshare]
pub struct Bar {
    pub value: String,
}

#[derive(Debug, Deserialize, Serialize)]
#[typeshare]
pub struct Dummy {
    /// generates a normal nested definition
    pub foo: Foo,
    #[serde(flatten)]
    /// generates a flattened definition
    pub bar: Bar,
}
```

then

```sh
typeshare . --lang=typescript --output-file=definitions.ts`
```

becomes

```ts
export interface Foo {
	value: string;
}

export interface Bar {
	value: string;
}

export interface Dummy extends Bar {
	/** generates a normal nested definition */
	foo: Foo;
}
```

It also does not affect the other languages' implementations for `Language`,
but the presence of `#[serde(flatten)]` does produce an expected `panic!` on e.g.

```sh
typeshare . --lang=swift --output-file=definitions.swift
```

A workaround could be to allow the consumer to define a blank `feature` flag,
and maybe `#[cfg_attr]` can be parsed to allow basically to ignore it on languages which does not (yet or never) support it, e.g.:

```rs
#[derive(Debug, Deserialize, Serialize)]
#[typeshare]
pub struct Dummy {
    /// generates a normal nested definition
    pub foo: Foo,
    #[cfg_attr(feature = "typescript", serde(flatten)])] // <--- this line
    /// generates a flattened definition
    pub bar: Bar,
}
```

But yes ultimately it's not super ergonomic, so looking for your feedback :)

Lucretiel:
Thanks for the PR! I'll give it a review next week

Lucretiel:
Sorry for the delay, reviewing now. really like the idea of using inheritance to solve this, but I'm concerned that this solution won't have an equivalent in languages that don't have inheritance (like Go), or that do but for which it's an unusual pattern (like Swift). I'll give it a review as written but I'll also do some discussion with the typeshare team at our next team meeting.

core/src/parser.rs
```diff
@@ -78,7 +78,7 @@ pub enum ParseError {
 }
 
 /// Parse the given Rust source string into `ParsedData`.
-pub fn parse(input: &str) -> Result<ParsedData, ParseError> {
```
Lucretiel, line 81: `supports_flatten` shouldn't be something the parser needs to worry about; it should unconditionally emit the fully parsed data (including `#[serde(flatten)]` attributes), and then that parsed data can be rejected by individual languages when the output is generated, if necessary.

Roms1383:
Alright! I understand your concern, please keep me posted in case I miss the discussion ! 👌

core/src/parser.rs
Roms1383: Sounds fair, but concretely just wondering how you would see it implemented.

core/src/language/typescript.rs
```diff
@@ -110,13 +110,33 @@ impl Language for TypeScript {
 
     fn write_struct(&mut self, w: &mut dyn Write, rs: &RustStruct) -> std::io::Result<()> {
         self.write_comments(w, 0, &rs.comments)?;
+        let mut inheritance = "".to_string();
+        let mut count = 0;
+        for field in rs.fields.iter() {
+            if field.flattened {
+                let ts_ty = match self.format_type(&field.ty, rs.generic_types.as_slice()) {
+                    Ok(v) => v,
+                    Err(e) => return Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, e)),
```
Lucretiel, line 119: Use `.map_err` and `?` here, instead of a `match`

core/src/language/typescript.rs
```diff
@@ -110,13 +110,33 @@ impl Language for TypeScript {
 
     fn write_struct(&mut self, w: &mut dyn Write, rs: &RustStruct) -> std::io::Result<()> {
         self.write_comments(w, 0, &rs.comments)?;
+        let mut inheritance = "".to_string();
+        let mut count = 0;
+        for field in rs.fields.iter() {
+            if field.flattened {
+                let ts_ty = match self.format_type(&field.ty, rs.generic_types.as_slice()) {
```
Lucretiel, line 117: `rs.generic_types.as_slice()` can probably be `&rs.generic_types`

core/src/language/typescript.rs
Roms1383: done in 9aa71fb

core/src/language/typescript.rs
Roms1383: done in c70b536

Einliterflasche:
Hey, is there any update on this feature, any specifics that are blocking it? I'd be willing to help.

Issue #22, opened 2022-11-24
Macro params to override generated type


fluxxu:
Thanks for the tool!
Currently this struct
```rust
#[typeshare]
pub struct OrderItem {
  #[ts(type = "any | null")]
  pub meta: Option<Value>,
}
```
will generate this ts type
```typescript
export interface OrderItem {
	meta?: Value;
}
```
where `Value` is `serde_json::Value`
I wonder if it's a good idea to support something like this
```rust
#[typeshare]
pub struct OrderItem {
  #[typeshare(type = "any | null")]
  pub meta: Option<Value>,
}
```

InquisitivePenguin:
Thanks for opening this issue! I like the idea behind this, but we'd need some way to specify that an override is for a particular language.

Alfred-Mountfield:
Would it be possible to just have something like `#[typeshare(js_type = "...", swift_type = "...", kotlin_type = "...")`?

I think an escape hatch like this is essential, you inevitably run into issues with combinations of `serde` tags, especially when you start dealing with `serde(flatten)`. My experience with these libraries are unfortunately TypeScript focused, but I hope Kotlin and Swift don't complicate it too much, some escape hatches in other libraries includes: 

**[Tsify's attribute macro](https://github.com/madonoharu/tsify)**
```rust
use tsify::Tsify;

#[derive(Tsify)]
pub struct Foo {
    #[tsify(type = "0 | 1 | 2")]
    x: i32,
}
```

**[`ts-rs`'s attribute macro](https://github.com/madonoharu/tsify)**
```rust
use ts_rs::TS;

#[derive(TS)]
#[ts(export)]
pub struct Foo {
    #[ts(type = "0 | 1 | 2")]
    x: i32,
}
```

InquisitivePenguin:
Hi @Alfred-Mountfield, and sorry for the wait! I'm going to be implementing a solution starting on Monday.

HHogg:
It would be really handy to have this, a case that I have is `chrono::NaiveDateTime` being serialized as a `string` but typeshare is still outputting as a `NaiveDateTime`. 

```rust
#[derive(Serialize)]
#[typeshare]
pub struct MyStruct {
  timestamp: NaiveDateTime,
}
```

```typescript
export interface MyStruct {
  timestamp: NaiveDateTime;
}
```

Would be really handy to do something like ...

```rust
pub struct MyStruct {
  #[typeshare(typescript = "string")]
  timestamp: NaiveDateTime,
[...]

PR #118, opened 2023-05-18
Field decorator parsing rework

InquisitivePenguin:
Needed for #22.

Parsing for `typeshare(<lang>(<decorator>))` only supports word decorators like `readonly` currently. This PR reworks how we parse field decorators to support name-value decorators too, like `typeshare(swift(type = "String"))`.

### DIFF
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -1,3 +1,4 @@
+use crate::rust_types::FieldDecorator;
 use crate::{
     language::SupportedLanguage,
     rename::RenameExt,
@@ -7,6 +8,7 @@ use crate::{
     },
 };
 use proc_macro2::{Ident, Span};
+use std::collections::BTreeSet;
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom,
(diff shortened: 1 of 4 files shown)


core/src/parser.rs
```diff
@@ -593,11 +590,18 @@ fn get_field_decorators(attrs: &[syn::Attribute]) -> HashMap<SupportedLanguage,
                 language,
                 list.iter()
                     .flat_map(|nested| match nested {
```
Lucretiel, line 592: Use `filter_map` instead of `flat_map` when the output type is `Option`

core/src/parser.rs
```diff
@@ -593,11 +590,18 @@ fn get_field_decorators(attrs: &[syn::Attribute]) -> HashMap<SupportedLanguage,
...
-                        NestedMeta::Meta(Meta::Path(path)) => path.get_ident(),
+                        NestedMeta::Meta(Meta::Path(path)) => {
```
Lucretiel, line 593: Recommend being more strict with the matching / acceptance here. Currently, this will accept things like `typeshare(a::b::c)` and return them as `FieldDecorator::Word("c")`.

core/src/rust_types.rs
```diff
@@ -75,8 +75,17 @@ pub struct RustField {
...
+    pub decorators: HashMap<SupportedLanguage, BTreeSet<FieldDecorator>>,
+}
+
+/// A single decorator on a field in Rust code.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum FieldDecorator {
+    /// A boolean flag enabled by its existence as a decorator: for example, `readonly`.
+    Word(String),
+    /// A key-value pair, such as `type = "any"`.
+    NameValue(String, String),
```
Lucretiel, line 88: Recommend using `syn` types like `Ident` instead of strings here

core/src/parser.rs
```diff
@@ -593,11 +590,18 @@ fn get_field_decorators(attrs: &[syn::Attribute]) -> HashMap<SupportedLanguage,
...
+                            Some(FieldDecorator::Word(path.get_ident()?.to_string()))
+                        }
+                        NestedMeta::Meta(Meta::NameValue(name_value)) => {
+                            Some(FieldDecorator::NameValue(
+                                name_value.path.get_ident()?.to_string(),
+                                literal_as_string(name_value.lit.clone())?,
+                            ))
+                        }
                         _ => None,
```
Lucretiel, line 602: It's likely that this branch should return an error instead of silently failing, since it indicates a malformed `#[typeshare]` attribute. This can be left to a future date, though.

core/src/rust_types.rs
InquisitivePenguin: We can't do this because parsing is multi-threaded and `Ident` is not `Send`able.

Lucretiel: Why on god's green earth is `Ident` not `Send`

Lucretiel (approved):

PR #119, opened 2023-05-19
Parameter to override a field type on a per-language basis

### LINKED ISSUE #22: Macro params to override generated type

InquisitivePenguin:
Resolves #22.

There is now a field decorator that can override the type of the field in the generated language(s) of your choice. This override takes precedent over everything else, including type mapping.


InquisitivePenguin:
This PR depends on #118.

core/data/tests/can_override_types/input.rs
```diff
@@ -0,0 +1,19 @@
+#[typeshare]
+#[serde(rename_all = "camelCase")]
+struct OverrideStruct {
+    // These annotations are intentionally inconsistent across languages
+    #[typeshare(swift(type = "Int"), typescript(readonly, type = "any | undefined"), kotlin(type = "Int"), go(type = "uint"), scala(type = "Short"))]
```
Lucretiel, line 5: Nit: recommend splitting these up over multiple lines

Lucretiel:
General questions:

- Is this the first `typeshare` attribute (aside from the top-level one)?
- Is there an assumption here that the types should be compatible without any additional (de)serialize logic, in either direction, in either rust or the target language?

Lucretiel (changes requested):

InquisitivePenguin:
> General questions:
> 
>     * Is this the first `typeshare` attribute (aside from the top-level one)?
> 
>     * Is there an assumption here that the types should be compatible without any additional (de)serialize logic, in either direction, in either rust or the target language?

* No, we already have field decorators like `#[typeshare(typescript(readonly))]`.
* Correct, we don't check for compatibility. That's the responsibility of the end-user (given that this is _usually_ a last-resort option)

Lucretiel (approved):

Lucretiel:
Slightly screwed up merging #118 so I force-rebased this branch. Merge should be fine.

PR #125, opened 2023-06-12
Add typescript option to generate readonly types

Diggsey:
It's common in typescript to avoid in-place mutations. This PR allows fully read-only types to be generated.

For object types, this could previously already be achieved via lots of annotations, but this is tedious and there was no way to apply the "readonly" modifier to maps and arrays.


Lucretiel:
I'm not sure I follow; we already support read-only types via the annotation. Is the idea that this is a global override that *forces* all types to be readonly, regardless of annotations?

Diggsey:
As I mentioned above, the annotations only allow you to make object properties read only. They don't allow you to make eg `readonly string[]`.

PR #127, opened 2023-06-19
Support tuple struct without fields

### LINKED ISSUE #42: Support for tuple structs without fields
Hi,

This is related to https://github.com/1Password/typeshare/issues/41.

The following:

```rust
#[typeshare::typeshare]
pub struct MyType();
```

...results in `index out of bounds: the len is 0 but the index is 0`

The following compiles, however:

```rust
#[typeshare::typeshare]
pub struct MyType(pub ());
```

czocher:
Handles tuple structs without any fields or visibility modifiers.

Closes #42

### DIFF
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -1,4 +1,4 @@
-use crate::rust_types::FieldDecorator;
+use crate::rust_types::{FieldDecorator, SpecialRustType};
 use crate::{
     language::SupportedLanguage,
     rename::RenameExt,
@@ -202,6 +202,16 @@ fn parse_struct(s: &ItemStruct) -> Result<RustItem, ParseError> {
             if f.unnamed.len() > 1 {
                 return Err(ParseError::ComplexTupleStruct);
             }
+
+            if f.unnamed.is_empty() {
+                return Ok(RustItem::Alias(RustTypeAlias {
+                    id: get_ident(Some(&s.ident), &s.attrs, &None),
+                    r#type: RustType::Special(SpecialRustType::Unit),
+                    comments: parse_comment_attrs(&s.attrs),
+                    generic_types,
+                }));
+            }
+
             let f = &f.unnamed[0];
 
             let ty = if let Some(ty) = get_field_type_override(&f.attrs) {
--- a/core/data/tests/can_handle_tuple_struct_without_fields/input.rs
+++ b/core/data/tests/can_handle_tuple_struct_without_fields/input.rs
@@ -1,0 +1,5 @@
+#[typeshare::typeshare]
+pub struct MyTypeA();
+
+#[typeshare::typeshare]
+pub struct MyTypeB(pub ());
--- a/core/data/tests/can_handle_tuple_struct_without_fields/output.go
+++ b/core/data/tests/can_handle_tuple_struct_without_fields/output.go
@@ -1,0 +1,8 @@
+package proto
+
+import "encoding/json"
+
+type MyTypeA struct{}
+
+type MyTypeB struct{}
+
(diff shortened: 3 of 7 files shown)


core/src/parser.rs
```diff
@@ -202,6 +202,16 @@ fn parse_struct(s: &ItemStruct) -> Result<RustItem, ParseError> {
             if f.unnamed.len() > 1 {
                 return Err(ParseError::ComplexTupleStruct);
             }
+
+            if f.unnamed.is_empty() {
+                return Ok(RustItem::Alias(RustTypeAlias {
```
Lucretiel, line 207: Shouldn't this be basically the same as a Unit struct?

czocher: It is, I just needed an early return ;). If you have a suggestion how to move things around to remove the duplication I'm open for changes!

Lucretiel: Sorry, I more mean that Unit structs are `RustItem::Struct`, but this is using a `RustItem::Alias`

czocher: I can change that, but then the representation of:
```
#[typeshare::typeshare]
pub struct MyType();
```

and 

```
#[typeshare::typeshare]
pub struct MyType(pub ());
```

would differ.

Currently
```
#[typeshare::typeshare]
pub struct MyType(pub ());
```

results in:

```
typealias MyType = Unit
```

in Kotlin. If I change it, it'll result in:

```
@Serializable
object MyType
```

In my opinion those representations should be the same, we can change either one of them. Which one would you prefer?  I added those two cases into the snapshot test in the PR.

Lucretiel: Actually, that's a great catch– we need to make sure that we're also consistent with the format that `serde` will use during serialization and deserialization. It looks like serde treats all tuple structs with length `0|2+` as arrays:

```
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef7881f4313744aa43e5797a569ceb0c

Lucretiel: Actually, that's a great catch; we need to make sure that the chosen representation aligns with how serde will treat the type. It appears that serde treats empty tuple structs the same was as non-newtype tuple structs:

```rust
#[derive(serde::Serialize)]
struct Unit;

#[derive(serde::Serialize)]
struct Tup0();

#[derive(serde::Serialize)]
struct Tup1(i32);

#[derive(serde::Serialize)]
struct Tup2(i32, i32);

fn main() {
    println!("{}", serde_json::to_string(&Unit).unwrap());
    println!("{}", serde_json::to_string(&Tup0()).unwrap());
    println!("{}", serde_json::to_string(&Tup1(3)).unwrap());
    println!("{}", serde_json::to_string(&Tup2(4, 5)).unwrap());
}
```

```
null
[]
3
[4,5]
```

[playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef7881f4313744aa43e5797a569ceb0c)

There's probably not a useful way to handle this round trip through an array in the destination language (for the same reason we don't support complex tuple structs), so we have to close this for now. Any support for empty tuple structs would also have to include support for larger, 2+ tuple structs, and some brief experimenting revealed that this would be very challenging in some languages (Swift especially).

Thanks for working through this with us. Discoveries like this are valuable for ensuring we don't back ourselves into a corner by supporting rust type shapes without clear analogues in other languages.

Lucretiel:
See [this comment](https://github.com/1Password/typeshare/pull/127#discussion_r1256632655) for a discussion of why we elected to close this; basically, Rust treats empty tuple structs the same as larger complex tuple structs, which means they round trip through an array in JSON, and there isn't yet a good way to ensure we can support array -> struct conversions in other languages (this is why we don't currently support complex tuple structs, especially given our lack of a specific use case for them).

core/src/parser.rs
czocher: Thanks for the info. If you need any further help with this issue feel free to contact me.

PR #129, opened 2023-06-19
Force Rust 1.57 MSRV

### LINKED ISSUE #94: Solve recurring MSRV issues
We keep running into issues regarding failing to compile on MSRV (1.57 as of this writing). 

- [ ] Check CI
- [ ] Make sure default toolchains are enforced in dev environments?

czocher:
This PR forces the CI as well as local dev environments to use Rust `1.57`, which is the MSRV for this project.

Closes #94


czocher:
@CerulanLumina what do we do about the `clippy` errors for MSRV?

CerulanLumina:
Clippy on stable should be fine. Any MSRV issues should be caught by the rest of the toolchain, and it seems using an updated clippy should only allow the extra indirection we have in `topsort`.

Thanks for working this, we appreciate your contributions!

czocher:
I changed it to validate with `stable` but also added an `allow` so that it won't bother people on their local environments.

Lucretiel:
Thanks for the contribution! I'm worried about forcing *all* uses of this library to use 1.57, even local development builds (as opposed to using it only as part of a CI matrix). I'll ask the rest of the typeshare team and get back to you.

czocher:
@Lucretiel I implemented the dev change due to the task mentioned in #94, but if that's not necessary I can remove the local dev part of this PR ;).

Lucretiel:
I spoke with the team and apparently I was just mistaken, these changes look good! I'm giving it a final lookover then we should be able to merge today

.github/workflows/ci.yml
```diff
@@ -15,8 +15,6 @@ jobs:
     strategy:
       matrix:
         rust:
-          - stable
```
Lucretiel, line 18: Any reason to remove this?

czocher: Yes, I assumed the CI should test only the MSRV. This assumption may not be correct - should I revert this one?

Lucretiel: Yeah, I'd like CI to run on both MSRV and current stable :)

Lucretiel (approved):

Lucretiel:
Thanks!

PR #130, opened 2023-06-20
Migrate to `syn` version `2.0`

### LINKED ISSUE #120: Move to `syn` 2.0
As the rust ecosystem moves towards syn 2.0 (and in particular now that `serde` uses it), we'd like to follow suit to ensure we're not bloating dependecy trees by asking users to use both syn 1 and syn 2.

czocher:
This is my take on migrating `typeshare` to `syn` version `2.0`. This will allow to use the `serde-derive-internals` library in the future.

# TODO
- [x] Fix failing tests, currently the type override does not work due to the fact it uses the `type` keyword to indicate the new type. It would be great if we could rename it to `new_type` but that would be a major breaking change, so instead I need to write my own parser.
- [x] Deduplicate methods such as `get_typeshare_meta_items` and `get_serde_meta_items` etc.
- [x] Change some of the helper methods into extension traits, e.g. `expr_to_string`, `remove_dash_from_identifier`, `rename_all_to_case`

Closes #120

### DIFF
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -463,7 +463,7 @@ pub(crate) fn get_name_value_meta_items<'a>(
 }
 
 /// Returns all arguments passed into `#[{ident}(...)]` where `{ident}` can be `serde` or `typeshare` attributes
-pub(crate) fn get_meta_items(attr: &syn::Attribute, ident: &str) -> Vec<Meta> {
+fn get_meta_items(attr: &syn::Attribute, ident: &str) -> Vec<Meta> {
     if attr.path().is_ident(ident) {
         attr.parse_args_with(Punctuated::<Meta, Token![,]>::parse_terminated)
             .iter()
@@ -475,7 +475,7 @@ pub(crate) fn get_meta_items(attr: &syn::Attribute, ident: &str) -> Vec<Meta> {
     }
 }
 
-pub(crate) fn get_ident(
+fn get_ident(
     ident: Option<&proc_macro2::Ident>,
     attrs: &[syn::Attribute],
     rename_all: &Option<String>,
@@ -491,7 +491,7 @@ pub(crate) fn get_ident(
     Id { original, renamed }
 }
 
-pub(crate) fn rename_all_to_case(original: String, case: &Option<String>) -> String {
+fn rename_all_to_case(original: String, case: &Option<String>) -> String {
     match case {
         None => original,
         Some(value) => match value.as_str() {
@@ -508,12 +508,12 @@ pub(crate) fn rename_all_to_case(original: String, case: &Option<String>) -> Str
     }
 }
 
-pub(crate) fn serde_rename(attrs: &[syn::Attribute]) -> Option<String> {
+fn serde_rename(attrs: &[syn::Attribute]) -> Option<String> {
     get_name_value_meta_items(attrs, "rename", SERDE).next()
 }
 
 /// Parses any comment out of the given slice of attributes
-pub(crate) fn parse_comment_attrs(attrs: &[Attribute]) -> Vec<String> {
+fn parse_comment_attrs(attrs: &[Attribute]) -> Vec<String> {
     attrs
         .iter()
         .map(|attr| attr.meta.clone())
@@ -527,7 +527,7 @@ pub(crate) fn parse_comment_attrs(attrs: &[Attribute]) -> Vec<String> {
 }
 
 // `#[typeshare(skip)]` or `#[serde(skip)]`
-pub(crate) fn is_skipped(attrs: &[syn::Attribute]) -> bool {
+fn is_skipped(attrs: &[syn::Attribute]) -> bool {
     attrs.iter().any(|attr| {
         get_meta_items(attr, SERDE)
             .into_iter()
@@ -544,18 +544,18 @@ fn serde_attr(attrs: &[syn::Attribute], ident: &str) -> bool {
     })
 }
 
-pub(crate) fn serde_default(attrs: &[syn::Attribute]) -> bool {
+fn serde_default(attrs: &[syn::Attribute]) -> bool {
     serde_attr(attrs, "default")
 }
 
-pub(crate) fn serde_flatten(attrs: &[syn::Attribute]) -> bool {
+fn serde_flatten(attrs: &[syn::Attribute]) -> bool {
     serde_attr(attrs, "flatten")
 }
 
 /// Checks the struct or enum for decorators like `#[typeshare(typescript(readonly)]`
 /// Takes a slice of `syn::Attribute`, returns a `HashMap<language, BTreeSet<decorator>>`, where `language` is `SupportedLanguage`
 /// and `decorator` is `FieldDecorator`. Field decorators are ordered in a `BTreeSet` for consistent code generation.
-pub(crate) fn get_field_decorators(
+fn get_field_decorators(
     attrs: &[Attribute],
 ) -> HashMap<SupportedLanguage, BTreeSet<FieldDecorator>> {
     let languages: HashSet<SupportedLanguage> = SupportedLanguage::all_languages().collect();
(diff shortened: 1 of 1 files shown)


czocher:
This PR is ready for review @Lucretiel @InquisitivePenguin.

Lucretiel:
Thanks for the contribution!

I wanted to ask about this:

> currently the type override does not work due to the fact it uses the type keyword to indicate the new type. It would be great if we could rename it to new_type but that would be a major breaking change, so instead I need to write my own parser.

I'm not sure I follow where this problem is occurring, could you elaborate? If this is about the use of `type` as an identifier in the typeshare code itself, it can usually be circumvented by using `r#type`, the raw identifier syntax. I'm surprised that this would require the addition of *so much* duplicated parse code.

Lucretiel (changes requested):
Thanks for the contribution!

I wanted to ask about this:

currently the type override does not work due to the fact it uses the type keyword to indicate the new type. It would be great if we could rename it to new_type but that would be a major breaking change, so instead I need to write my own parser.

I'm not sure I follow where this problem is occurring, could you elaborate? If this is about the use of type as an identifier in the typeshare code itself, it can usually be circumvented by using r#type, the raw identifier syntax. I'm surprised that this would require the addition of so much duplicated parse code.

czocher:
> I'm not sure I follow where this problem is occurring, could you elaborate? If this is about the use of type as an identifier in the typeshare code itself, it can usually be circumvented by using r#type, the raw identifier syntax. I'm surprised that this would require the addition of so much duplicated parse code.

@Lucretiel let me explain, cause I see the misunderstanding here. This PR consists of 3 commits:
1. Extracting the helper functions to `helper.rs` and implementing `syn 2.0`.
2. Removing the duplicated methods between the `serde` annotation parsing and `typeshare` annotation parsing (which after the change were the same, the only difference being the name of the annotation)
3. A final cleanup which consisted of migrating the large `parser.rs` file into  a `parser` module an multiple, much smaller files, each with a separate responsibility.

Here's my motivation:
While migrating the `syn` crate version I noticed that the `parser.rs` file structure was rather messy. It dealt with two separate things: parsing and helper functions (it even had a `// helpers` comment on [line 420](https://github.com/1Password/typeshare/pull/130/files#diff-539b0ae2e9fc1061a52b9a8584af6cca7d84f61cd3691c63df352ffeacbdd48cL420), which immediately told me it needed to be divided into two files), it had a lot of duplication. 
I decided to extract all the parsing helpers to `helpers.rs` as a first pass, and later implement the other refactorings in two separate PRs, so we can easily revert them if you found it too big of a change.

The part about the `type`, which I mentioned in my first `TODO` task was due to the fact that `type` is a keyword in Rust (as you already know). `syn 1` didn't treat it in any special way, but `syn 2` failed parsing and required me to use a different method to parse it as an identifier (`Ident::parse_any` instead of just the default `Ident::parse`). [This](https://github.com/1Password/typeshare/pull/130/files#diff-f186e048cc77157ee4b866a0b7bd8f4e2cf29f0b9301b61cb3e516904bb3ff64R193) is the part I needed to implement as my own parser. If that wasn't the case I could just use: `attr.parse_args_with(Punctuated::<Meta, Token![,]>::parse_terminated)` instead of that whole closure.

The rest is mostly the same code, just migrated to the new `syn` version and extracted to more, smaller files for readability ;).

Lucretiel:
> A final cleanup which consisted of migrating the large parser.rs file into a parser module and multiple, much smaller files, each with a separate responsibility.

Ah, GitHub was hiding this part from me, it looked like you were adding just a huge amount of new code (cause it made unclear that the old `parser.rs` was deleted. Re-reviewing with this in mind.

Lucretiel (changes requested):
It looks like `syn` 1 is still coming in from a few places; it looks like they're all trivial to fix:

- `typeshare-annotation`: mostly a no-op, should be a trivial fix
- `serde`: bump to at least `1.0.157`, though probably just go to latest for some recent build speed improvements
- `thiserror`: bump to `1.0.40`

Lucretiel:
(for reference, you can use `cargo tree --invert --package syn@1.0.100` to track down where these dependencies are coming from)

czocher:
> Ah, GitHub was hiding this part from me, it looked like you were adding just a huge amount of new code (cause it made unclear that the old `parser.rs` was deleted. Re-reviewing with this in mind.

My fault, should have described my motivations better from the get-go ;).

czocher:
Updated the dependencies, this task should be mergable @Lucretiel

core/src/parser/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+mod enum_parser;
+mod error;
+/// Helpers used throughout the parsing procedure.
+pub mod helpers;
+mod struct_parser;
+mod type_alias_parser;
+pub use error::ParseError;
+
+/// Trait used to parse the input file and provide our internal representations for various Rust constructs.
+pub trait Parser<'a> {
```
Lucretiel, line 16: I'd like to remove this trait and just use inherent methods on the relevant types. I'm not seeing a need for the trait, since there isn't any functionality that's generic over some `T: Parser`.

core/src/parser/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+            syn::Item::Enum(item_enum) if has_typeshare_annotation(&item_enum.attrs) => {
+                RustEnum::parse(item_enum).map(Some)
+            }
+            syn::Item::Type(item_type) if has_typeshare_annotation(&item_type.attrs) => {
+                RustTypeAlias::parse(item_type).map(Some)
+            }
+            _ => Ok(None),
+        }
+    }
+}
```
Lucretiel, line 107: Could you partially revert this to be more similar to the original? My concern is mostly with this flow of `RustItem` -> `ParsedData containing a single item` -> `collect into ParsedData`. This involves a lot of unnecessary allocations, and more semantically it round trips through these types in unnecessary ways.

Lucretiel (changes requested):
Sorry for the additional delays. I'd like to revert the parser changes and re-review them in a different PR, since they involve implementation changes, rather than being a direct verbatim copy of the existing code from `parser.rs`.

Additionally, run `cargo update -p wasm-bindgen` in the project root; this will remove `syn 1.0` entirely from the `Cargo.lock`.

Thanks again for all your work on this!

core/src/parser/mod.rs
Lucretiel: Actually, because this refactoring includes implementation changes rather than a verbatim copy of the old items into the new submodules, I'd like to ask that you revert these changes here and re-submit them as a new PR so that they can be more specifically reviewed.

czocher:
@Lucretiel I reverted the commit but left the helpers extraction (can revert that too if you deem it necessary). Feel free to do a review.

czocher:
The review fixes has been implemented and I believe this PR is ready to be merged @Lucretiel.

andymac4182:
@Lucretiel Any update on this? Its conflicting now due to the length of time.

czocher:
I'm still around to do a rebase whenever it's necessary. Probably gonna do one soon just to make it easier in the future.

andymac4182:
@InquisitivePenguin @CerulanLumina Any thoughts on this from you?

czocher:
Rebased it.

anish-1p:
@czocher Thanks for all your work on this! Could we also get the helpers extraction reverted as well, so the PR exclusively contains changes related to the migration to `syn 2`. We'd be happy to take a look at the helpers extraction in a separate PR.

czocher:
@anish-1p the helper extraction has been reverted.

anish-1p (approved):
This looks good to me. Only change I'd request is to make some utility functions in `core/src/parser.rs` private (I don't see a reason why `get_meta_items`, `get_ident`, `rename_all_to_case`, `is_skipped`, `serde_default`, `serde_flatten` and `get_decorators` are public to the crate).

czocher:
@anish-1p I made the functions private.

Lucretiel (approved):

PR #131, opened 2023-06-21
Support Rust slice type

### LINKED ISSUE #126: Typeshare panics when encountering slice of user type.
For example, a struct like so:
```rust
#[typeshare]
#[derive(Serialize)]
pub struct Video<'a> {
    pub tags: &'a [Tag],
}
```
Running typeshare gives an error like so:
```
thread 'main' panicked at 'failed to parse a rust type: Unexpected token when parsing type: `[Tag]`. This is an internal error, please ping a typeshare developer to resolve this problem.', /home/louisgariepy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/typeshare-cli-1.6.0/src/main.rs:267:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`Tag` is a simple enum which is also annotated with  `#[typeshare]`.

According to the error message, the error happens because of the slice type `[Tag]`.

Note that changing `&'a [Tag]` for `&'a Vec<Tag>` makes things work.

czocher:
Adds support for the Rust slice type e.g.:

```rust
#[typeshare]
#[derive(Serialize)]
pub struct Video<'a> {
    pub tags: &'a [Tag],
}
```

@LouisGariepy please verify if it works in your case.

Closes #126

### DIFF
--- a/core/data/tests/can_generate_slice_of_user_type/input.rs
+++ b/core/data/tests/can_generate_slice_of_user_type/input.rs
@@ -1,0 +1,5 @@
+#[typeshare]
+#[derive(Serialize)]
+pub struct Video<'a> {
+    pub tags: &'a [Tag],
+}
--- a/core/data/tests/can_generate_slice_of_user_type/output.go
+++ b/core/data/tests/can_generate_slice_of_user_type/output.go
@@ -1,0 +1,7 @@
+package proto
+
+import "encoding/json"
+
+type Video struct {
+	Tags []Tag `json:"tags"`
+}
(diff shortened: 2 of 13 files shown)


LouisGariepy:
@czocher Tested your branch on my project and it seems to work as intended as far as I can tell. Thanks.

Lucretiel:
This looks great! In the future, if it comes up, we can look at adding a warning for `&[u8]`, which may not do what the user expects (both serde and typeshare will treat it as an array of integers rather than a byte blob), but we can leave that for if people have issues with it.

Lucretiel (approved):

PR #140, opened 2023-08-26
Allow overrides for disallowed types (e.g. 64 bit integers)

tomjw64:
Here's an alternative approach to support 64 bit integer types than the one taken in https://github.com/1Password/typeshare/pull/112.

Rather than blessing one approach, with https://github.com/1Password/typeshare/pull/119 being merged, it seems like it's possible to accept this oddity of TypeScript/JavaScript and force users of Typeshare (or rather, only those using it to output TypeScript) to make a decision appropriate for them and their use-case.


tomjw64:
More or less resolves https://github.com/1Password/typeshare/issues/24

wyatt-herkamp:
I swapped my install to this this fork. It is working.

tomjw64:
@Lucretiel @InquisitivePenguin Sorry to bother, but could I get a review on this?

annie444:
This closes #112 for sure. A very beautiful solution I might add!

mattyg:
I'd like to see this included as well. 

You may also want to add u128 and i128 to the supported types.

mattyg:
@anish-1p sorry to bother you -- would you mind taking a look at this PR and reviewing or merging? Thank you!

esmevane:
@tomjw64 Thanks for putting this together - I've just followed a trail of breadcrumbs here and you've helped me out a lot!

bwilliams-sequence:
@charlespierce can you review this?  This is a blocker for many of us working with 64 bits ints.

tomjw64:
@charlespierce Thanks for the review!

Could these both be resolved by also requiring type overrides for i64 and u64 for the other languages? That should be a requirement that could be relaxed in the future while maintaining backwards-compatibility.

charlespierce:
Yeah, updating the other languages to also require an override for `usize` & `isize` would definitely be a reasonable solution!

tomjw64:
@charlespierce Updated to also require overrides for all languages for pointer sized types and updated docs. LMK if you would like any other changes!

Issue #35, opened 2022-12-15
Support for generating TypeScript union types for non-algebraic enums 


Alfred-Mountfield:
In a lot of situations, plain union types are better suited than enums. Ideally there would be some way to make the output of this:
```rust
#[derive(Serialize, Deserialize)]
#[typeshare]
pub enum FooBar {
    Foo,
    Bar,
}
```

be this:
```typescript
export type FooBar = "Foo" | "Bar"
```

instead of 
```typescript
export enum FooBar {
    Foo = "Foo",
    Bar = "Bar",
}
```

I had hoped I'd be able to coerce the generator into doing as desired by using the serde `tag` attributes:
 `#[serde(untagged)]` is unsupported

`#[serde(tag = "type", content = "content")]` doesn't work due to
```
thread '<unnamed>' panicked at 'the serde tag attribute is not supported for non-algebraic enums: FooBar', /Users/foo/.cargo/registry/src/github.com-1ecc6299db9ec823/typeshare-cli-1.0.1/src/main.rs:234:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If I pollute my Rust code with meaningless newtypes:
```Rust
#[derive(Serialize, Deserialize)]
#[typeshare]
#[serde(tag = "type", content = "content")]
pub enum FooBar {
    Foo(()),
    Bar(()),
}
```

Then I do get a TypeScript type that I can then manipulate, but this Rust pollution is not ideal.
```typescript
export type FooBar = 
	| { type: "Foo", content: undefined }
	| { type: "Bar", content: undefined };
```

where I can then
```typescript
import { FooBar as FooBarGen } from "./foobar.gen";

export type FooBar = FooBarGen["type"] // resolves to "Foo" | "Bar"
```

CerulanLumina:
Hello, and thank you for the suggestion! I'll bring this up with the other team members to discuss how we'd like to implement this.

Alfred-Mountfield:
Great, I appreciate it! For what it's worth, I think a lot of issues like this could have their impact mitigated heavily by some solution to #22, most type generation libraries I've seen (although they are focused on TypeScript) all provide escape hatches for the weird edge-cases you'll inevitably run into with what `serde` can let you do, I'll comment directly on that issue though.

stewartadam:
@LuminaSapphira any updates on moving this forward or reviewing #145? Its rather unfortunate to have to use newtypes or empty named fields to get the variants to export properly.

PR #145, opened 2023-09-27
Allow to generate unions instead of TS enums in Typescript

akozhemiakin:
Addresses #35

Now for unit enums, you can choose to generate a TypeScript union type instead
of a TypeScript enum.

```rust
#[typeshare(ts_union)]
pub enum UnitEnum {
    VariantA,
    VariantB,
    VariantC,
}
```

This would generate the following TypeScript code:
```ts
export type UnitEnum = "VariantA" | "VariantB" | "VariantC";
```

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -34,7 +34,7 @@ The test suite can of course be run normally without updating any expectations:
 cargo test -p typeshare-core
 ```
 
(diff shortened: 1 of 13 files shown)


Tehnix:
Any chance of reviving this? If there's interest in getting this merged I wouldn't mind taking a look at it

PR #142, opened 2023-09-04
Draft implementation of algebraic enum without `content` attribute

FlorianDenis:
Draft implementation of support for Algebraic enums without specifying the `content` attribute (see https://github.com/1Password/typeshare/issues/141).

Only Unit variants and anonymous structs variants are supported (tuple variants explicitly require the content attribute), and I built support for Kotlin, Scala, Swift and TypeScript output.

This was done by adding a new `RustEnum` type (`AlgebraicFlattened`, the name is not great, I welcome any other suggestion) without the `content_key`.

PR #146, opened 2023-10-02
Add support for internally tagged non-tuple enums for Typescript

akozhemiakin:
Currently, generating types for internally tagged enums is not possible, even though serde [supports it](https://serde.rs/enum-representations.html#internally-tagged).

This PR allows the omission of the "content" attribute while still raising errors for languages other than TypeScript. Support for additional languages may be introduced later if necessary and feasible. Additionally, it triggers an error when attempting to skip the "content" attribute for tuple enum variants in TypeScript.


akozhemiakin:
Unfortunately, I completely missed https://github.com/1Password/typeshare/pull/142. The implementations are different, but https://github.com/1Password/typeshare/pull/142 appears to already support more than just TypeScript. I'll close this for now, feel free to reopen it if needed.

PR #154, opened 2024-02-01
Pass decorators to anonymous struct generators

### LINKED ISSUE #149: Add the ability to apply decorators to anonymous struct enum variants
Consider the following enum decorated with the swift protocol `Equatable`.

```rust
#[typeshare(swift = "Equatable")]
enum MyEnum {
    MyVariant {
        a: String,
        b: f64
    }
}
```

Running this through typeshare, the following is generated

```swift
/// Generated type representing the anonymous struct variant `MyVariant` of the `MyEnum` Rust enum
public struct MyEnumMyVariantInner: Codable {
	public let a: String
	public let b: Double

	public init(a: String, b: Double) {
		self.a = a
		self.b = b
	}
}
public enum MyEnum: Codable, Equatable {
	case myVariant(MyEnumMyVariantInner)
       
        ...
}
```

Although `Equatable` was applied to `MyEnum`, it was not applied to `MyEnumMyVariantInner`. And there doesn't seem to be any way to do so.

My first thought was to apply `#[typeshare(swift = "Equatable")]` on the enum variant, but that was ignored.

```diff
#[typeshare(swift = "Equatable")]
#[serde(tag = "type", content = "content")]
enum MyEnum {
+  #[typeshare(swift = "Equatable")]
    MyVariant {
        a: String,
        b: f64
    }
}
```

czocher:
Closes #149.

### DIFF
--- a/core/data/tests/algebraic_enum_decorator/input.rs
+++ b/core/data/tests/algebraic_enum_decorator/input.rs
@@ -31,3 +31,12 @@ pub enum BestHockeyTeams4 {
     PittsburghPenguins,
     Lies(String),
 }
+
+#[typeshare(swift = "Equatable")]
+#[serde(tag = "type", content = "content")]
+enum MyEnum {
+    MyVariant {
+        a: String,
+        b: f64
+    }
+}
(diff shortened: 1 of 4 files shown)


charlespierce:
Hi @czocher, it looks like the core underlying change from this PR was actually already added in #170, so I believe this should be working with the next release!

Issue #158, opened 2024-03-28
Add support for prefixing Kotlin type names


smohammad:
Add a new argument to set an optional prefix value that is used to prefix type names in the generated Kotlin file. This is already available for Swift using `swift-prefix` and would be available in the same way but for Kotlin types.

PR #159, opened 2024-04-01
Support prefixing Kotlin type names

smohammad:
Supports prefixing Kotlin type names with a new `kotlin-prefix` argument that can be supplied when using the typeshare tool.

Resolves https://github.com/1Password/typeshare/issues/158/


DrewBurkhart (commented):
LGTM

rudyrichter (approved):
These look consistent with what we did on the swift side

PR #162, opened 2024-04-02
Fix mdbook github action

darrell-roberts:
We probably don't have to use the MSRV for the github mdbook action. Right now it is broken and this fixes it. `actions-rs` was also failing and I noticed the project is [archived](https://github.com/actions-rs). Switching to [`actions-rust-lang`](https://github.com/actions-rust-lang/setup-rust-toolchain) with the changes below gets the action running correctly once again.


smohammad (approved):
Changes look good to me!

PR #168, opened 2024-04-23
Remove Kotlin compose dependency and star imports

vrendina:
This PR makes two minor changes to the generated Kotlin code:

- The usage `androidx.compose.runtime.NoLiveLiterals` creates an unnecessary dependency on the Jetpack Compose runtime in our generated types module. It also isn't needed because there are no literals in the generated output.
- Usage of wildcard/star imports is discouraged. Updated to import the specific serialization dependencies we use to be consistent with Kotlin best practices.


smohammad (approved):
Looks good to me!

PR #169, opened 2024-04-30
Add Typeshare support for Python

hculea:
This PR is adding initial support for Python in Typeshare (gated by a feature flag, like Go). 

See [@MOmarMiraj 's comment](https://github.com/1Password/typeshare/pull/169#issuecomment-2450523731) for a high-level description of the design considerations behind this work.

This was made possible thanks to the prior work of @adriangb: https://github.com/1Password/typeshare/pull/25.


core/Cargo.toml
```diff
@@ -14,6 +14,9 @@ thiserror = "1"
 itertools = "0.12"
 lazy_format = "2"
 joinery = "2"
+topological-sort = { version = "0.2.2"}
+once_cell = { version = "1"}
```
Lucretiel, line 18: Use the built-in oncelock https://doc.rust-lang.org/std/sync/struct.OnceLock.html

core/src/language/python.rs
```diff
@@ -0,0 +1,779 @@
...
+            .cloned()
+            .collect();
+        missing.sort();
+        res.extend(missing);
+        Ok(res)
+    }
+}
+
+#[derive(Debug, Clone)]
+enum ParsedRusthThing<'a> {
```
Lucretiel, line 100: Typo: parsedrustthing

core/src/language/python.rs
```diff
@@ -0,0 +1,779 @@
...
+    v.retain(|e| uniques.insert(e.clone()));
+}
+
+/// All information needed to generate Python type-code
+#[derive(Default)]
+pub struct Python {
+    /// Mappings from Rust type names to Python type names
+    pub type_mappings: HashMap<String, String>,
+    /// The Python module for the generated code.
+    pub module: RefCell<Module>,
```
Lucretiel, line 160: Why a refcell here?

core/src/language/python.rs
```diff
@@ -0,0 +1,779 @@
...
+            let mut module = self.module.borrow_mut();
+            for alias in &data.aliases {
+                let thing = ParsedRusthThing::TypeAlias(alias);
+                let identifier = self.get_identifier(thing);
+                match &alias.r#type {
+                    RustType::Generic { id, parameters: _ } => {
+                        module.add_global(identifier, vec![id.clone()])
+                    }
+                    RustType::Simple { id } => module.add_global(identifier, vec![id.clone()]),
+                    RustType::Special(_) => {}
```
Lucretiel, line 184: Type aliases for special types are unsupported?

core/src/language/python.rs
```diff
@@ -0,0 +1,779 @@
...
+            | SpecialRustType::U32
+            | SpecialRustType::I54
+            | SpecialRustType::U53 => Ok("int".into()),
+            SpecialRustType::F32 | SpecialRustType::F64 => Ok("float".into()),
+            SpecialRustType::Bool => Ok("bool".into()),
+            SpecialRustType::U64
+            | SpecialRustType::I64
+            | SpecialRustType::ISize
+            | SpecialRustType::USize => {
+                panic!("64 bit types not allowed in Typeshare")
```
Lucretiel, line 314: They're supported by python, though, so I'm not seeing any reason to fail here.

Lucretiel:
Doing some review, sorry for the delay. I'd like to ask for a high-level description in this MR for how this integration works, with especial focus on how structs are handled (in particular how the JSON <-> struct boundary is handled).

MOmarMiraj:
Forgot to mention but we currently do not support generics as we have yet to figure out on an ideal/idiomatic way to handle them in Python so we left it out of this PR.

hculea:
Thanks for the mention @MOmarMiraj 😄 That should be okay while Python support is still experimental (FWIW, Go doesn't support generics either)

hculea:
@Lucretiel does @MOmarMiraj 's comment above answer to your high-level description request? Otherwise, we're happy to provide more context on the choices that were made here. Thanks!

core/data/tests/test_algebraic_enum_case_name_support/output.py
```diff
@@ -0,0 +1,55 @@
...
+    pass
+
+class AdvancedColorsTypes(str, Enum):
+    STRING = "string"
+    NUMBER = "number"
+    NUMBER_ARRAY = "number-array"
+    REALLY_COOL_TYPE = "reallyCoolType"
+
+class AdvancedColors(BaseModel):
+    model_config = ConfigDict(use_enum_values=True)
```
Lucretiel, line 21: Any particular reason to use `use_enum_values`? Can the pydantic->JSON layer handle that conversion for us?

core/data/tests/test_generate_char/output.py
```diff
@@ -0,0 +1,12 @@
+"""
+ Generated by typeshare 1.12.0
+"""
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+
+class MyType(BaseModel):
+    field: str
```
Lucretiel, line 10: Consider `constr` to constrain the length here, since it's known to be a `char` https://docs.pydantic.dev/1.10/usage/types/#arguments-to-constr

core/data/tests/test_optional_type_alias/output.py
```diff
@@ -0,0 +1,19 @@
...
+
+OptionalU16 = int
+
+
+OptionalU32 = int
+
+
+class FooBar(BaseModel):
+    foo: OptionalU32
+    bar: OptionalU16
```
Lucretiel, line 17: This looks very wrong to me– the relevant rust type is `Option<u32>`

core/data/tests/use_correct_integer_types/output.py
```diff
@@ -0,0 +1,20 @@
...
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+
+class Foo(BaseModel):
+    """
+    This is a comment.
+    """
+    a: int
```
Lucretiel, line 13: Nit: I don't feel nearly as strongly about this as I do about by other feedback, but if you wanted to be fancy, you could use `conint` to handle fixed-width integers with a min and max value: https://docs.pydantic.dev/1.10/usage/types/#arguments-to-conint

core/src/language/python.rs
```diff
@@ -0,0 +1,836 @@
...
+
+use super::CrateTypes;
+
+use convert_case::{Case, Casing};
+// Collect unique type vars from an enum field
+// Since we explode enums into unions of types, we need to extract all of the generics
+// used by each individual field
+// We do this by exploring each field's type and comparing against the generics used by the enum
+// itself
+fn collect_generics_for_variant(variant_type: &RustType, generics: &[String]) -> Vec<String> {
```
Lucretiel, line 23: Nit: consider using a wrapper that pushes to an `&mut Vec<String>` here, to avoid the deep recursive frequent allocations and data copying between vectors. Also consider returning a `HashSet` or `BTreeSet` from this function.

Lucretiel (changes requested):
Mostly this looks good. I had some concerns about algebraic enum representations, along with a few edge cases (mostly around serde renames) that appear to be handled incorrectly. Rust side of this looks fine, if quite allocation-heavy; I'll leave it be for now but there's a lot of low-hanging fruit for improvement if typeshare-python needs to be faster.

Lucretiel:
Follow-up: it looks like most of the pydantic `con*` types (`conlist`, `constr`, etc) are deprecated, in favor of `Field` and `Annotated`, so substitute all my recommendations as appropriate.

Lucretiel:
As a subsequent follow up, one idea for a way to handle unions would resemble this:

```
class MyEnumVariant1:
    type: Literal["Variant1"]
    content: MyEnumVariant1Content

...

MyEnum = Union[MyEnumVariant1, ...]
```

I'm still a novice around pydantic, so there could certainly be a better way that I'm not aware of. For all I know this pattern is fully built-in to pydantic somewhere and I just don't know where.

core/data/tests/can_apply_prefix_correctly/output.py
```diff
@@ -0,0 +1,74 @@
...
+    NUMBER = "Number"
+    NUMBER_ARRAY = "NumberArray"
+    REALLY_COOL_TYPE = "ReallyCoolType"
+    ARRAY_REALLY_COOL_TYPE = "ArrayReallyCoolType"
+    DICTIONARY_REALLY_COOL_TYPE = "DictionaryReallyCoolType"
+
+class AdvancedColors(BaseModel):
+    model_config = ConfigDict(use_enum_values=True)
+    t: AdvancedColorsTypes
+    c: Union[Dict[str, ItemDetailsFieldValue], ItemDetailsFieldValue, List[ItemDetailsFieldValue], List[int], int, str]
```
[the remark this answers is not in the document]
CheatCod: It is deterministic (in the sort order of `std::collections::BTreeSet` which I'm pretty sure is some kind of alphanumeric order). I think it is confusing to the user as the variant classes themselves are written in declaration order. I changed to a `IndexSet` which preserves insertion order!

core/data/tests/anonymous_struct_with_rename/output.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class AnonymousStructWithRenameKebabCase(BaseModel):
+    """
+    Generated type representing the anonymous struct variant `KebabCase` of the `AnonymousStructWithRename` Rust enum
+    """
+    model_config = ConfigDict(populate_by_name=True)
+
+    another_list: List[str] = Field(alias="another-list")
+    camel_case_string_field: str
```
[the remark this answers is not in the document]
CheatCod: I believe you are correct and the codegen is not handling renames correctly. The correct output in this case should be:

```rs
another_list: List[str]
camelCaseStringField : str
```

core/data/tests/generate_types/output.py
```diff
@@ -0,0 +1,27 @@
...
+class Types(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
+    s: str
+    static_s: str
+    int_8: int
+    float: float
+    double: float
+    array: List[str]
+    fixed_length_array: List[str]
```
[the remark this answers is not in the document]
CheatCod: Will be using `Annotated` with `Field` instead.

core/data/tests/test_generate_char/output.py
CheatCod: Will be using Annotated with Field instead.

CheatCod:
> As a subsequent follow up, one idea for a way to handle unions would resemble this:
> 
> ```
> class MyEnumVariant1:
>     type: Literal["Variant1"]
>     content: MyEnumVariant1Content
> 
> ...
> 
> MyEnum = Union[MyEnumVariant1, ...]
> ```
> 
> I'm still a novice around pydantic, so there could certainly be a better way that I'm not aware of. For all I know this pattern is fully built-in to pydantic somewhere and I just don't know where.

After some experimentation and discussion with @MOmarMiraj, I think this is the best approach for [externally tagged enums](https://serde.rs/enum-representations.html#externally-tagged). I've refactored the PR to do so.

core/data/tests/anonymous_struct_with_rename/output.py
```diff
@@ -0,0 +1,72 @@
...
+
+class AnonymousStructWithRenameTypes(str, Enum):
+    LIST = "list"
+    LONG_FIELD_NAMES = "longFieldNames"
+    KEBAB_CASE = "kebabCase"
+
+class AnonymousStructWithRename(BaseModel):
+    model_config = ConfigDict(use_enum_values=True)
+    type: AnonymousStructWithRenameTypes
+    content: Union[AnonymousStructWithRenameKebabCase, AnonymousStructWithRenameList, AnonymousStructWithRenameLongFieldNames]
```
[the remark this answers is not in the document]
CheatCod: I have decided to go with your suggestion below for [externally tagged enums](https://serde.rs/enum-representations.html#externally-tagged)

core/data/tests/can_apply_prefix_correctly/output.py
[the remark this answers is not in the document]
Lucretiel: Presumably this will actually be superceded by my proposal for tagged unions further down

core/data/tests/anonymous_struct_with_rename/output.py
```diff
@@ -0,0 +1,57 @@
...
+    something_else: bool = Field(alias="something-else")
+
+
+class AnonymousStructWithRenameTypes(str, Enum):
+    LIST = "list"
+    LONG_FIELD_NAMES = "longFieldNames"
+    KEBAB_CASE = "kebabCase"
+
+class AnonymousStructWithRenameList(BaseModel):
+    type: AnonymousStructWithRenameTypes = AnonymousStructWithRenameTypes.LIST
```
Lucretiel, line 46: It's possible I'm misunderstanding pydantic, but shouldn't this be a `Literal` parameter? I had understood this `=` to just set a default value. 

https://docs.pydantic.dev/2.0/usage/types/unions/

core/data/tests/test_algebraic_enum_case_name_support/output.py
Lucretiel: Understood, thanks for the explaination.

core/data/tests/test_optional_type_alias/output.py
CheatCod: Fixed

core/data/tests/anonymous_struct_with_rename/output.py
CheatCod: Fixed

core/data/tests/use_correct_integer_types/output.py
CheatCod: I can't resolve this thread for some reason. Marking as won-fix to prevent scope-creeping.

core/data/tests/test_generate_char/output.py
CheatCod: Marking as won-fix to prevent scope-creeping.

CheatCod:
For future improvement, let me know if I've missed anything.

1. `conint` from pydantic can be used for further validation. For example, if the Rust type is u8, then we can leverage `conint` to bound the number from 0 to 255.

2. `struct` and `enum` with generic parameters are not supported. For example, the test case `can_generate_generic_enum` is not enabled for Python as it is not supported.

3. Rust `enum` must be adjacently tagged, like `#[serde(tag = "t", content = "c")]`, other enum representations, such as internally tagged and untagged are not support. Read more about enum representation with serde here: https://serde.rs/enum-representations.html. Correct me if I am wrong but I believe this is a limitation in Typeshare itself

core/Cargo.toml
hculea: This has been done, and the import removed 😄

hculea:
Thanks @CheatCod! I've captured these in https://github.com/1Password/typeshare/issues/217 to make sure we don't lose track of them once merging this PR. I've resolved all the outstanding conversations after making sure they have either been logged as a future improvement or actually resolved.   

@Lucretiel I think this should be ready to go! Let me know if you have any other questions or concerns 👍

darrell-roberts (commented):
Left some comments about multiple file generation support.

core/src/language/python.rs
```diff
@@ -0,0 +1,784 @@
...
+    }
+
+    fn write_imports(
+        &mut self,
+        _writer: &mut dyn Write,
+        _imports: super::ScopedCrateTypes<'_>,
+    ) -> std::io::Result<()> {
+        todo!()
+    }
+}
```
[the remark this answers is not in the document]
hculea: I've added multi-file support to the list of follow-ups here, and linked your solution: https://github.com/1Password/typeshare/issues/217

I expect that we will have a use-case for multi-file support in the new year, with the introduction of multiple release channels in the SDK. So we will likely then prioritize these effort.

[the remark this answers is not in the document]
hculea: I've now made this a no-op and linked the issue regarding the limitations in `README.md`.

core/src/language/python.rs
```diff
@@ -0,0 +1,784 @@
...
+    /// Whether or not to exclude the version header that normally appears at the top of generated code.
+    /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
+    pub no_version_header: bool,
+}
+
+impl Language for Python {
+    fn type_map(&mut self) -> &HashMap<String, String> {
+        &self.type_mappings
+    }
+    fn generate_types(
```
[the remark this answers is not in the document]
hculea: Linked in the follow-up issue! Thanks for the suggestion!

Lucretiel (approved):
Sorry I didn't approve before I left for the long weekend!

PR #170, opened 2024-04-30
Decorator to redact Kotlin toString

jonduran3000:
Adds a Kotlin-language specific decorator in order to generate a custom `toString` function on data classes where all the field data removed. This is a guard to avoid potentially leaking sensitive data to logs via the default `toString` function.


srriddle (changes requested):
Add a test for the redacted_to_string, but otherwise looks good to me.

srriddle (approved):

srriddle (approved):

PR #172, opened 2024-05-08
Fix typeshare skip

darrell-roberts:
I found that the `#[typeshare(skip)]` attribute on a field does not work with  anonymous structs in an Enum.


core/src/parser.rs
```diff
@@ -385,6 +385,7 @@ fn parse_enum_variant(
             fields: fields_named
                 .named
                 .iter()
+                .filter(|f| !is_skipped(&f.attrs))
```
savannidgerinel, line 388: Please add an automated test case. Otherwise, this looks good.

savannidgerinel (approved):
Ran this against a test project. It worked as predicted.

CerulanLumina (approved):

smohammad (approved):

core/src/parser.rs
darrell-roberts: Added snapshot test.

PR #173, opened 2024-06-04
Allow swift decorator constraints to apply to generics

darrell-roberts:
Allow `#[typeshare(swift = "Constraint")]` to be applied to generic members.

Example in included test case.

### DIFF
--- a/core/data/tests/generic_struct_with_constraints_and_decorators/input.rs
+++ b/core/data/tests/generic_struct_with_constraints_and_decorators/input.rs
@@ -1,0 +1,17 @@
+#[typeshare(swift = "Equatable")]
+pub struct Button<T> {
+    /// Label of the button
+    pub label: String,
+    /// Accessibility label if it needed to be different than label
+    pub accessibility_label: Option<String>,
+    /// Optional tooltips that provide extra explanation for a button
+    pub tooltip: Option<String>,
+    /// Button action if there one
+    pub action: Option<T>,
+    /// Icon if there is one
+    pub icon: Option<Icon>,
+    /// Button state
+    pub state: ButtonState,
+    /// Button Mode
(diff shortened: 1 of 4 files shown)


kevinbhayes:
Will we ever want to apply different decorators to the generics? i.e. 
`#[typeshare(swift = "Constraint", swiftGeneric = "DifferentConstraint")]`

charlespierce (approved):
This change LGTM! I'm not familiar with Swift, but would we need to do similar logic in `write_enum`? [This block](https://github.com/1Password/typeshare/blob/2262a56a93f9dd41f133f733607d2c8947fdc87a/core/src/language/swift.rs#L449-L452) seems to be doing essentially the same as what you replaced in `write_struct`.

darrell-roberts:
> This change LGTM! I'm not familiar with Swift, but would we need to do similar logic in `write_enum`? [This block](https://github.com/1Password/typeshare/blob/2262a56a93f9dd41f133f733607d2c8947fdc87a/core/src/language/swift.rs#L449-L452) seems to be doing essentially the same as what you replaced in `write_struct`.

Yup I'll update that as well.

kevinbhayes:
Update: yes, there is a case where we don't want to apply the same decorators to both the struct and its generic type. We often want structs to be Identifiable, but do not want to place that restriction on the generic type. We should look at a mechanism to define separate decorators for the struct itself, and the generic type.

kevinbhayes:
@charlespierce unfortunately, this can break any code with generics that don't need the same constraints as the struct itself. e.g. the generic type doesn't need Identifiable, but the struct does.

@darrell-roberts is working on a fix to allow separate designation of generic constraint decorators, so please coordinate with him on getting the fix in soon, or consider rolling back this one.

Thanks.

PR #174, opened 2024-06-05
allow defining generic type constraints for swift

darrell-roberts:
This updates how generic type constraints are applied to swift types. Before if a type had a generic member we would implicitly apply the same type constraints to the generic constraints. That is not always desirable. This allows the user to define the generic type constraints via the `swiftGenericConstraints` typeshare attribute.

See example in `core/data/tests/generic_struct_with_constraints_and_decorators/input.rs`.

Also removed excessive allocations in swift.rs.

Introduced swift config option `codablevoid_constraints` which allows defining constraints for `CodableVoid`.

### DIFF
--- a/core/src/language/swift.rs
+++ b/core/src/language/swift.rs
@@ -242,8 +242,7 @@ impl Language for Swift {
         self.write_comments(w, 0, &ty.comments)?;
 
         let swift_prefix = &self.prefix;
-        let type_name =
-            swift_keyword_aware_rename(Cow::Owned(format!("{}{}", swift_prefix, ty.id.renamed)));
+        let type_name = swift_keyword_aware_rename(format!("{}{}", swift_prefix, ty.id.renamed));
 
         writeln!(
             w,
@@ -266,8 +265,7 @@ impl Language for Swift {
         writeln!(w)?;
         self.write_comments(w, 0, &rs.comments)?;
 
-        let type_name =
-            swift_keyword_aware_rename(Cow::Owned(format!("{}{}", self.prefix, rs.id.renamed)));
+        let type_name = swift_keyword_aware_rename(format!("{}{}", self.prefix, rs.id.renamed));
 
         // If there are no decorators found for this struct, still write `Codable` and default decorators for structs
         // Check if this struct's decorators contains swift in the hashmap
@@ -310,9 +308,7 @@ impl Language for Swift {
             if f.id.renamed.chars().any(|c| c == '-') {
                 coding_keys.push(format!(
                     r##"{} = "{}""##,
-                    remove_dash_from_identifier(
-                        swift_keyword_aware_rename(Cow::Borrowed(&f.id.renamed)).as_ref()
-                    ),
+                    remove_dash_from_identifier(swift_keyword_aware_rename(&f.id.renamed).as_ref()),
                     &f.id.renamed
                 ));
 
@@ -321,7 +317,7 @@ impl Language for Swift {
                 should_write_coding_keys = true;
             } else {
                 coding_keys.push(remove_dash_from_identifier(
-                    swift_keyword_aware_rename(Cow::Borrowed(&f.id.renamed)).as_ref(),
+                    swift_keyword_aware_rename(&f.id.renamed).as_ref(),
                 ));
             }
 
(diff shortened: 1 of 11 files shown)


core/src/language/swift.rs
```diff
@@ -812,6 +787,67 @@ impl Swift {
...
+    /// it with the `default_generic_constraints` configuration. If no `swiftGenericConstraints` is defined then we just use
+    /// `default_generic_constraints`.
+    fn generic_constraints<'a>(
+        &'a self,
+        decorator_map: &'a DecoratorMap,
+        generic_types: &'a [String],
+    ) -> String {
+        let swift_generic_contraints_annotated = decorator_map
+            .get(&SupportedLanguage::Swift)
+            .and_then(|d| d.get(SWIFT_GENERIC_CONSTRAINTS_DECORATOR))
```
charlespierce, line 801: It feels a bit awkward here (and elsewhere, especially in the `kotlin` section) to have to first get the values for a specific language, and then drill down into the decorators that match the kind we are looking for. Now that we have multiple decorators for a single language (and we already don't support decorators for _every_ language), would it be unreasonable to decouple the decorator type from the `SupportedLanguage` struct?

If we had a `DecoratorKind` enum that included the supported decorators (e.g. `Kotlin`, `Swift`, and `SwiftGenericConstraints`), and then built the map of `DecoratorKind` => `Vec` of values, the API would be roughly similar to what it is now, but allowing multiple kinds of decorators to be pulled when they are relevant.

darrell-roberts: Yup, I agree. Pushed the update.

core/src/language/swift.rs
```diff
@@ -803,20 +792,71 @@ impl Swift {
...
 }
 
-fn swift_keyword_aware_rename(name: &str) -> String {
-    if SWIFT_KEYWORDS.contains(&name) {
-        return format!("`{}`", name);
+fn swift_keyword_aware_rename(name: Cow<str>) -> Cow<str> {
+    if SWIFT_KEYWORDS.contains(&name.as_ref()) {
+        Cow::Owned(format!("`{name}`"))
+    } else {
+        name
```
charlespierce, line 860: Returning `Cow<str>` here makes a lot of sense, however needing to pass one in strikes me as a bit of an anti-pattern. It looks like that might be used to avoid lifetime issues with the `format!` calls above and "value dropped while borrowed" or something? If so, would it be reasonable to take an `Into<Cow>` to avoid needing to update all of the call sites? I'm thinking something like:

```rust
fn swift_keyword_aware_rename<'a, T>(name: T) -> Cow<'a, str>
where
    T: Into<Cow<'a, str>>,
{
    let name = name.into();
    if SWIFT_KEYWORDS.contains(&name.as_ref()) {
        Cow::Owned(format!("`{name}`")
    } else {
        name
    }
}
```

That way we still avoid the extra allocations, but we can directly pass either a `String` or `&str` into the method without needing to explicitly annotate `Cow::Owned` or `Cow::Borrowed` at the call site.

core/src/language/swift.rs
```diff
@@ -261,51 +266,38 @@ impl Language for Swift {
...
+            Either::Left(
+                self.get_default_decorators().chain(
+                    swift_decs
+                        .iter()
+                        .filter(|d| d.as_str() != CODABLE)
+                        .map(|s| s.as_str()),
+                ),
+            )
+        } else {
+            Either::Right(self.get_default_decorators())
```
charlespierce, line 287: This is a nice pattern! I don't think I've seen `Either` used like this for the two branches to allow a simple iterator fold at the end, but it's a lot more readable than collecting into an intermediate vector.

charlespierce (approved):
Overall LGTM! Left one suggestion for cleaning up the need to pass `Cow` into a helper, but that's not blocking. I'm still waiting on some more info about the release process, so I'll look to merge this in the morning. If you get to making that change by then, great; if not, happy to have it as a follow-up PR since it doesn't really affect the working of the module, just the dev ergonomics.

core/src/language/swift.rs
darrell-roberts: Yup I felt I missed something when repeatedly wrapping in `Cow`. Pushed update.

Issue #63, opened 2023-02-10
Support target-specific sharing of types


InquisitivePenguin:
Taken from our internal issue tracker.

### Original issue description

As of today, writing something like the following doesn't work:

```
#[cfg_attr(target_vendor = "apple", typeshare)]
```
It would be nice to support that, or, more generally, the concept of typesharing a type for a specific target only.

InquisitivePenguin:
I agree with the original issue but I think the solution would work better as part of the `typesahre` attribute itself, instead of an external attribute that wraps `typeshare`.

PR #176, opened 2024-06-07
Allow skipping fields/variants via target_os argument

darrell-roberts:
When defining types you may not want to typeshare all types/fields/variants to all target platforms. Rust allows you to restrict  definitions via the `#[cfg(target_os = "platform"]` macro. 

This PR allows typeshare to run with an optional command line argument `--target-os`. When checking for what fields/variants to skip, typeshare will see if this argument is set and compare check if it matches any `#[cfg(target_os)]` attributes. If the argument and attribute target do not match then it will be skipped. Similar to `#[typeshare(skip)]` but based on the `--target-os` argument not matching.

Full file skipping is supported as well when using `#![cfg(target_os = "platform_os")]` as well. Not supported yet is attributes on `mod` declarations.

See snapshot test in `core/data/tests/excluded_by_target_os`.

This is a possible solution for https://github.com/1Password/typeshare/issues/63

### DIFF
--- a/core/data/tests/excluded_by_target_os_full_module/input.rs
+++ b/core/data/tests/excluded_by_target_os_full_module/input.rs
@@ -1,7 +1,7 @@
 #![cfg(feature = "online")]
 #![allow(dead_code)]
 #![cfg(any(target_os = "android", feature = "testing"))]
-#![cfg(target_os = "android")]
+#![cfg(target_os = "wasm32")]
 
 #[typeshare]
 pub struct IgoredUnlessAndroid;
--- a/core/data/tests/excluded_by_target_os_full_module/input.rs
+++ b/core/data/tests/excluded_by_target_os_full_module/input.rs
@@ -1,0 +1,7 @@
+#![cfg(feature = "online")]
+#![allow(dead_code)]
+#![cfg(any(target_os = "android", feature = "testing"))]
+#![cfg(target_os = "android")]
+
+#[typeshare]
+pub struct IgoredUnlessAndroid;
--- a/core/data/tests/excluded_by_target_os_full_module/input.rs
+++ b/core/data/tests/excluded_by_target_os_full_module/input.rs
@@ -4,4 +4,4 @@
 #![cfg(target_os = "wasm32")]
 
 #[typeshare]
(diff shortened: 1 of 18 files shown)


core/data/tests/excluded_by_target_os_full_module/input.rs
```diff
@@ -0,0 +1,7 @@
+#![cfg(feature = "online")]
+#![allow(dead_code)]
+#![cfg(any(target_os = "android", feature = "testing"))]
+#![cfg(target_os = "wasm32")]
+
+#[typeshare]
+pub struct IgoredUnlessAndroid;
```
Progdrasil, line 7: Nit: 
```suggestion
pub struct IgnoredUnlessAndroid;
```

Progdrasil (approved):

sanjanadesai27 (approved):
Overall these changes all look good to me, though I'm not an expert in Typeshare just yet. 
A couple of minor improvements I could suggest include: 
The code could benefit from additional doc comments explaining the purpose and usage of the new `--target-os` argument and the logic behind skipping fields, variants, and entire files based on the `#[cfg(target_os = "platform")]` attribute. Additionally, I think it would make sense to split out helper functions like `target_os_skip` and `target_os_from_meta_list` in a separate file just to keep the code organized and the files concise. 

But, I'd consider these suggestions non-blocking and overall I think this is in solid state.

sanjanadesai27 (approved):
Overall these changes all look good to me, though I'm not an expert in Typeshare just yet.
A couple of minor improvements I could suggest include:
The code could benefit from additional doc comments explaining the purpose and usage of the new --target-os argument and the logic behind skipping fields, variants, and entire files based on the #[cfg(target_os = "platform")] attribute. Additionally, I think it would make sense to split out helper functions like target_os_skip and target_os_from_meta_list in a separate file just to keep the code organized and the files concise.

But, I'd consider these suggestions non-blocking and overall I think this is in solid state.

complexspaces (approved):

complexspaces:
This has been released as part of [1.10.0-beta.6](https://crates.io/crates/typeshare-cli/1.10.0-beta.6) on crates.io.

PR #182, opened 2024-07-15
Add support for Kotlin inline value classes

jonduran3000:
There are instances where we are using a type alias in Kotlin when we should be using an [inline value class](https://kotlinlang.org/docs/inline-classes.html). This pull request adds a Kotlin-specific typeshare decorator for generating these inline value classes. Additionally, updated the `redacted_to_string` decorator to `Redacted` so there is a consistent naming convention for the language-specific decorators.

### DIFF
--- a/core/data/tests/struct_decorator/input.rs
+++ b/core/data/tests/struct_decorator/input.rs
@@ -15,7 +15,7 @@ pub struct BestHockeyTeams2 {
     Lies: String,
 }
 
-#[typeshare(kotlin = "Redacted")]
+#[typeshare(redacted)]
 pub struct BestHockeyTeams3 {
     PittsburghPenguins: u32,
     Lies: String,
@@ -27,5 +27,5 @@ pub struct BestHockeyTeams4 {
     Lies: String,
 }
 
-#[typeshare(kotlin = "JvmInline, Redacted")]
+#[typeshare(kotlin = "JvmInline", redacted)]
 pub struct BestHockeyTeams5(String);
--- a/core/data/tests/struct_decorator/input.rs
+++ b/core/data/tests/struct_decorator/input.rs
@@ -15,7 +15,7 @@ pub struct BestHockeyTeams2 {
     Lies: String,
 }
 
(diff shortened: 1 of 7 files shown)


core/src/language/kotlin.rs
```diff
@@ -119,16 +120,50 @@ impl Language for Kotlin {
...
-                .unwrap_or_default(),
-            self.format_type(&ty.r#type, ty.generic_types.as_slice())
-                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?
-        )?;
+        if self.is_inline(&ty.decorators) {
+            writeln!(w, "@Serializable")?;
+            writeln!(w, "@JvmInline")?;
+            writeln!(w, "value class {}{}(", self.prefix, ty.id.renamed)?;
+
+            self.write_element(
```
Lucretiel, line 128: Is this block equivalent to a similar block in the `write_struct` code?

Lucretiel:
Wondering if the redacted string code can be deduplicated; otherwise this looks great to me

Lucretiel (approved):

core/src/language/kotlin.rs
jonduran3000: The main difference between this and the one in `write_struct` is that structs are being translated into `data` [classes](https://kotlinlang.org/docs/data-classes.html) (or an `object` if it contains no data) and this is specifically a `value` [class](https://kotlinlang.org/docs/inline-classes.html).

core/data/tests/struct_decorator/input.rs
```diff
@@ -15,7 +15,7 @@ pub struct BestHockeyTeams2 {
     Lies: String,
 }
 
-#[typeshare(kotlin = "redacted_to_string")]
+#[typeshare(kotlin = "Redacted")]
```
darrell-roberts, line 18: I think the other languages can also have redacted output. It may be better for this to be parsed as `#[typeshare(redacted)]` allowing each language formatter to choose how the redacted type should be implemented.

darrell-roberts:
LGTM

PR #184, opened 2024-07-17
Remove requirement of -go-package from typeshare with go feature

MOmarMiraj:
Currently the typeshare CLI with go feature enabled, doesn't allow the usage of package in `typeshare.toml` as the CLI requires the --go-package to be there. This MR will tackle that issue by not requiring the --go-package.

I also added a check to ensure that the package name is passed as this is a requirement in Golang. 

I also added a UT for reading the package name from a `typeshare.toml`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -232,7 +232,7 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> anyhow::R
         .get_many::<String>(ARG_TARGET_OS)
         .map(|arg| arg.into_iter().map(ToString::to_string).collect::<Vec<_>>())
         .unwrap_or_default();
-    config
+    Ok(config)
 }
 
 /// Prints out all parsing errors if any and returns Err.
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -10,6 +10,7 @@ use args::{
 use clap::ArgMatches;
 use config::Config;
 use ignore::{overrides::OverrideBuilder, types::TypesBuilder, WalkBuilder};
+use log::error;
 use parse::{all_types, parse_input, parser_inputs};
 use rayon::iter::ParallelBridge;
 use std::collections::{BTreeMap, HashMap};
@@ -30,6 +31,11 @@ mod parse;
 mod writer;
 
 fn main() -> anyhow::Result<()> {
+    flexi_logger::Logger::try_with_env()
+        .unwrap()
+        .start()
+        .unwrap();
+
     #[allow(unused_mut)]
     let mut command = build_command();
 
@@ -122,7 +128,7 @@ fn main() -> anyhow::Result<()> {
         parser_inputs(walker_builder, language_type, multi_file).par_bridge(),
         &ignored_types,
         multi_file,
-        target_os,
+        &target_os,
     )?;
 
     // Collect all the types into a map of the file name they
@@ -222,9 +228,11 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> anyhow::R
         }
         assert_go_package_present(&config)?;
     }
-
-    config.target_os = options.value_of(ARG_TARGET_OS).map(|s| s.to_string());
-    Ok(config)
+    config.target_os = options
+        .get_many::<String>(ARG_TARGET_OS)
+        .map(|arg| arg.into_iter().map(ToString::to_string).collect::<Vec<_>>())
+        .unwrap_or_default();
+    config
 }
 
 /// Prints out all parsing errors if any and returns Err.
@@ -236,7 +244,7 @@ fn check_parse_errors(parsed_crates: &BTreeMap<CrateName, ParsedData>) -> anyhow
     {
         errors_encountered = true;
         for error in &data.errors {
-            eprintln!(
+            error!(
                 "Parsing error: \"{}\" in crate \"{}\" for file \"{}\"",
                 error.error, error.crate_name, error.file_name
             );
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -217,14 +217,10 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> anyhow::R
 
     #[cfg(feature = "go")]
     {
-        match options.value_of(args::ARG_GO_PACKAGE) {
-            Some(go_package) => {
-                config.go.package = go_package.to_string();
-            }
-            None => {
-                config.go.package = is_go_package_present(&config)?;
-            }
+        if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
+            config.go.package = go_package.to_string();
         }
+        assert_go_package_present(&config)?;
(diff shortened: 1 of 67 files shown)


cli/src/main.rs
```diff
@@ -40,7 +40,7 @@ fn main() -> anyhow::Result<()> {
                 .long("go-package")
                 .help("Go package name")
                 .takes_value(true)
-                .required_if(ARG_TYPE, "go"),
+                .required(false),
```
hculea, line 43: What would happen if the package name is specified neither in the flag nor in the config file?

MOmarMiraj: Similar to the other languages, it would come without a package name and this would error out in Go. I have updated it so theres a check to see if package name is given through the flags or `typeshare.toml`. If neither than error out.

cli/src/main.rs
```diff
@@ -246,3 +252,23 @@ fn check_parse_errors(parsed_crates: &HashMap<CrateName, ParsedData>) -> anyhow:
...
+
+    let config_file = options.value_of(ARG_CONFIG_FILE_NAME);
+    let config = config::load_config(config_file).context("Unable to read configuration file");
+    match config {
+        Ok(config) => {
+            if !config.go.package.is_empty() {
+                true
+            } else {
+                false
+            }
```
seanaye, line 270:
```suggestion
            !config.go.package.is_empty()
```

cli/src/main.rs
```diff
@@ -246,3 +252,23 @@ fn check_parse_errors(parsed_crates: &HashMap<CrateName, ParsedData>) -> anyhow:
         Ok(())
     }
 }
+
+#[cfg(feature = "go")]
+fn is_package_present(options: &ArgMatches) -> bool {
```
seanaye, line 257: would it be slightly more readable to have a function that returns `anyhow::Result<String>` and call it on main line 219?

`    if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {`

here we could error out if the go package is not defined

seanaye (changes requested):

cli/src/main.rs
MOmarMiraj: Edited the function so we check directly here.

cli/src/main.rs
```diff
@@ -40,11 +40,17 @@ fn main() -> anyhow::Result<()> {
...
         );
     }
 
     let options = command.get_matches();
+    // #[cfg(feature = "go")]
+    // {
+    //     if !is_package_present(&options) {
+    //         return Err(anyhow!("Please provide a package name in the typeshare.toml or using --go-package <package name>"));
+    //     }
+    // }
```
seanaye, line 53: leftover comments

cli/src/main.rs
```diff
@@ -216,12 +222,16 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> Config {
...
 
     #[cfg(feature = "go")]
-    if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
-        config.go.package = go_package.to_string();
+    {
+        if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
+            config.go.package = go_package.to_string();
+        } else {
+            config.go.package = is_go_package_present(&config)?;
+        }
```
seanaye, line 230: prefer match over if else
```suggestion
        match options.value_of(args::ARG_GO_PACKAGE) {
            Some(go_package) => {
                config.go.package = go_package.to_string();
            },
            None => {
                config.go.package = is_go_package_present(&config)?;
            }
        }
```

seanaye (changes requested):
2 small suggestions, otherwise LGTM, can approve after

cli/src/main.rs
MOmarMiraj: Done!

seanaye (approved):

cli/src/main.rs
```diff
@@ -216,12 +216,19 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> Config {
...
     #[cfg(feature = "go")]
-    if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
-        config.go.package = go_package.to_string();
+    {
+        match options.value_of(args::ARG_GO_PACKAGE) {
+            Some(go_package) => {
+                config.go.package = go_package.to_string();
+            }
+            None => {
+                config.go.package = is_go_package_present(&config)?;
```
seanaye, line 225: Setting this value is a little redundant because we are just reading from the config and then setting it again. I think it would make more sense to just do nothing on this line and then assert that the package exists

cli/src/main.rs
```diff
@@ -246,3 +253,13 @@ fn check_parse_errors(parsed_crates: &BTreeMap<CrateName, ParsedData>) -> anyhow
...
+
+#[cfg(feature = "go")]
+fn is_go_package_present(config: &Config) -> anyhow::Result<String> {
+    if !config.go.package.is_empty() {
+        return Ok(config.go.package.clone());
+    }
+    Err(anyhow!(
+        "Please provide a package name in the typeshare.toml or using --go-package <package name>"
+    ))
+}
```
seanaye, line 265:
```suggestion
fn assert_go_package_present(config: &Config) -> anyhow::Result<()> {
    if config.go.package.is_empty() {
        return Err(anyhow!(
            "Please provide a package name in the typeshare.toml or using --go-package <package name>"
        ));
    }
    Ok(())
}
```

cli/src/main.rs
```diff
@@ -216,12 +216,19 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> Config {
     }
 
     #[cfg(feature = "go")]
-    if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
-        config.go.package = go_package.to_string();
+    {
+        match options.value_of(args::ARG_GO_PACKAGE) {
+            Some(go_package) => {
+                config.go.package = go_package.to_string();
+            }
```
seanaye, line 223: since we would do nothing in the None branch this might be better expressed as an `if let` statement

e.g.
```rust
if let Some(go_package) = options.value_of(args::ARG_GO_PACKAGE) {
    config.go.package = go_package.to_string();
}
```

cli/src/main.rs
```diff
@@ -216,12 +216,19 @@ fn override_configuration(mut config: Config, options: &ArgMatches) -> Config {
...
+            None => {
+                config.go.package = is_go_package_present(&config)?;
+            }
+        }
```
seanaye, line 227:
```suggestion
            assert_go_package_is_present(&config)?;
        }
```

seanaye (changes requested):
Sorry to re-request changes but I think we could make this a little more idiomatic (and also avoid and extra string allocation)

cli/src/main.rs
MOmarMiraj: Done!

PR #187, opened 2024-07-31
Allow multiple target-os

darrell-roberts:
* Added support for multiple `--target-os` values. 
* Added support for handling `#[cfg(not(..))]` target_os parsing.
* Updated test in [core/data/tests/excluded_by_target_os/input.rs](https://github.com/darrell-roberts/typeshare/blob/multiple-target-os-option/core/data/tests/excluded_by_target_os/input.rs) to include all possible target_os cfg definitions.
* Added flexi_logger/replaced `println` `eprintln`.

To run the test and see the evaluation:

- `RUST_LOG=debug cargo t --all-features --test snapshot_tests excluded_by_target_os::typescript -- --nocapture`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Version 1.10.0-beta
 
-## 1.10.0-beta.7
+## 1.10.0-beta.8
 
 - Added support for [inline value classes](https://kotlinlang.org/docs/inline-classes.html) in Kotlin - [#182](https://github.com/1Password/typeshare/pull/182)
 - Added the ability to specify that a struct should have information redacted - [#170](https://github.com/1Password/typeshare/pull/170)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Version 1.10.0-beta
 
-## 1.10.0-beta.8
+## 1.10.0-beta.7
 
 - Added support for [inline value classes](https://kotlinlang.org/docs/inline-classes.html) in Kotlin - [#182](https://github.com/1Password/typeshare/pull/182)
 - Added the ability to specify that a struct should have information redacted - [#170](https://github.com/1Password/typeshare/pull/170)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,4 +36,3 @@ inherits = "release"
 [workspace.dependencies]
 log = "0.4"
 flexi_logger = "0.28"
-anyhow = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,10 @@ ci = ["github"]
 installers = ["shell", "powershell"]
 # Target platforms to build apps for (Rust target-triple syntax)
 targets = [
-    "x86_64-unknown-linux-gnu",
-    "x86_64-apple-darwin",
-    "x86_64-pc-windows-msvc",
-    "aarch64-apple-darwin",
+  "x86_64-unknown-linux-gnu",
+  "x86_64-apple-darwin",
+  "x86_64-pc-windows-msvc",
+  "aarch64-apple-darwin",
 ]
 
 [profile.test]
@@ -32,3 +32,8 @@ incremental = true
 # The profile that 'cargo dist' will build with
 [profile.dist]
 inherits = "release"
+
+[workspace.dependencies]
+log = "0.4"
+flexi_logger = "0.28"
+anyhow = "1"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -23,4 +23,6 @@ rayon = "1.10"
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
 typeshare-core = { path = "../core", version = "1.10.0-beta.7" }
-anyhow = "1"
+anyhow.workspace = true
+log.workspace = true
+flexi_logger.workspace = true
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "typeshare-cli"
-version = "1.10.0-beta.7"
+version = "1.10.0-beta.8"
 edition = "2021"
 description = "Command Line Tool for generating language files with typeshare"
 license = "MIT OR Apache-2.0"
@@ -22,7 +22,7 @@ once_cell = "1"
 rayon = "1.10"
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
-typeshare-core = { path = "../core", version = "1.10.0-beta.7" }
+typeshare-core = { path = "../core", version = "1.10.0-beta.8" }
 anyhow.workspace = true
 log.workspace = true
 flexi_logger.workspace = true
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "typeshare-cli"
-version = "1.10.0-beta.8"
+version = "1.10.0-beta.7"
 edition = "2021"
 description = "Command Line Tool for generating language files with typeshare"
 license = "MIT OR Apache-2.0"
@@ -22,7 +22,7 @@ once_cell = "1"
 rayon = "1.10"
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
-typeshare-core = { path = "../core", version = "1.10.0-beta.8" }
+typeshare-core = { path = "../core", version = "1.10.0-beta.7" }
 anyhow.workspace = true
 log.workspace = true
 flexi_logger.workspace = true
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -23,6 +23,6 @@ rayon = "1.10"
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
 typeshare-core = { path = "../core", version = "1.10.0-beta.7" }
-anyhow.workspace = true
 log.workspace = true
 flexi_logger.workspace = true
+anyhow = "1"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "typeshare-core"
-version = "1.10.0-beta.7"
+version = "1.10.0-beta.8"
 license = "MIT OR Apache-2.0"
 edition = "2021"
 description = "The code generator used by Typeshare's command line tool"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -7,6 +7,7 @@ description = "The code generator used by Typeshare's command line tool"
 repository = "https://github.com/1Password/typeshare"
 
 [dependencies]
+anyhow.workspace = true
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "2", features = ["full", "visit"] }
@@ -14,9 +15,10 @@ thiserror = "1"
 itertools = "0.12"
 lazy_format = "2"
 joinery = "2"
+log.workspace = true
+flexi_logger.workspace = true
 
 [dev-dependencies]
-anyhow = "1"
 expect-test = "1.5"
 once_cell = "1"
 cool_asserts = "2"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "typeshare-core"
-version = "1.10.0-beta.8"
+version = "1.10.0-beta.7"
 license = "MIT OR Apache-2.0"
 edition = "2021"
 description = "The code generator used by Typeshare's command line tool"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -7,7 +7,6 @@ description = "The code generator used by Typeshare's command line tool"
 repository = "https://github.com/1Password/typeshare"
 
 [dependencies]
-anyhow.workspace = true
 proc-macro2 = "1"
 quote = "1"
 syn = { version = "2", features = ["full", "visit"] }
@@ -23,3 +22,4 @@ expect-test = "1.5"
 once_cell = "1"
 cool_asserts = "2"
 syn = { version = "2", features = ["full", "visit", "extra-traits"] }
+anyhow = "1"
--- a/core/data/tests/excluded_by_target_os/input.rs
+++ b/core/data/tests/excluded_by_target_os/input.rs
@@ -47,3 +47,15 @@ pub struct ManyStruct;
 #[typeshare]
 #[cfg(any(target_os = "android", target_os = "ios"))]
 pub struct MultipleTargets;
+
+#[typeshare]
+#[cfg(not(any(target_os = "android", target_os = "ios")))]
+pub struct DefinedTwice {
+    field1: u64,
+}
+
+#[typeshare]
+#[cfg(any(target_os = "android", target_os = "ios"))]
+pub struct DefinedTwice {
+    field1: String,
+}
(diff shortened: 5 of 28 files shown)


core/src/visitors.rs
```diff
@@ -266,44 +279,50 @@ impl<'ast, 'a> Visit<'ast> for TypeShareVisitor<'a> {
...
+    //     if let Some(target_os) = self.target_os.as_ref() {
+    //         if i.attrs.iter().any(|attr| target_os_skip(attr, target_os)) {
+    //             debug!("skip {}", i.ident);
+    //         }
+    //     };
 
-        syn::visit::visit_item_mod(self, i);
-    }
+    //     syn::visit::visit_item_mod(self, i);
+    // }
```
seanaye, line 322: leftover commented code

seanaye (approved):
Everything looks good, just some leftover comments

core/data/tests/excluded_by_target_os/input.rs
```diff
@@ -32,3 +44,35 @@ pub enum Test {}
 #[cfg(feature = "super")]
 #[cfg(target_os = "android")]
 pub enum SomeEnum {}
+
+#[typeshare]
+#[cfg(any(target_os = "ios", taget_os = "android"))]
+pub struct ManyStruct;
```
seanaye, line 50: It looks like this type is not in the output files due to the bug with the ordering of the `target_os`.

`MultipleTargets` type is included because "android" is the first listed target, but this type is currently being skipped because "android" is the second listed target.

seanaye (changes requested):
Just noticed this bug is still occurring while testing locally. It appears like its illustrated in the tests too

core/data/tests/excluded_by_target_os/input.rs
darrell-roberts: It actually works. The input in my test is wrong. It says `taget_os` instead of `target_os`. I'll update the test.

core/src/visitors.rs
darrell-roberts: It was there before but not really used. In the future I want to handle `#[cfg(target_os = "target")]` on `mod` definitions. It doesn't do that now but I'd like to add that later on.

core/Cargo.toml
```diff
@@ -7,16 +7,18 @@ description = "The code generator used by Typeshare's command line tool"
 repository = "https://github.com/1Password/typeshare"
 
 [dependencies]
+anyhow.workspace = true
```
charlespierce, line 10: For my own curiosity, was there something in these changes that necessitated moving `anyhow` to be a regular dependency, rather than a dev dependency?

Cargo.toml
```diff
@@ -32,3 +32,8 @@ incremental = true
 # The profile that 'cargo dist' will build with
 [profile.dist]
 inherits = "release"
+
+[workspace.dependencies]
+log = "0.4"
+flexi_logger = "0.28"
```
charlespierce, line 38: Would it be a huge burden to split the logging changes into a separate PR? It can be a bit difficult to review when there are multiple, unconnected changes in the same PR.

core/src/parser.rs
```diff
@@ -618,54 +636,115 @@ fn is_redacted(attrs: &[syn::Attribute]) -> bool {
     })
 }
 
-/// Check if we have a `target_os` cfg that dooes not match command line
-/// argument `--target-os`.
+/// Check if this attribute rejects our optional `target_os` arguments.
```
charlespierce, line 639: Admittedly I don't have a ton of experience with the `get_meta_items` and similar parsing here, however I find it a bit difficult to track what combinations are being checked through these methods. Could we add some comments (or add to the doc comments) describing what kind of expression each of these is parsing and how?

Related, given that `any`, `all`, `not`, etc. can be combined in basically any way, are there groups that we don't support? If so, should we document what settings do and do not work with typeshare?

charlespierce (commented):
Overall looks nice! Mostly comments about making sure we leave ourselves a good record of how things work so that going forward we can maintain / improve these areas.

core/Cargo.toml
darrell-roberts: Bot the `cli` and `core` crates are using it so any common dependency I'll move to the workspace decencies.

Cargo.toml
darrell-roberts: While developing these new use cases for `target_os` I needed to observe how I was evaluating each in detail and that required debug logging. I was putting `println` but these debug log statements were something I wanted to retain in case I need to look into what is going wrong in the future. I decided we should have proper logging so debug logging can be retained and enabled. So the reason I added the log dependency is for this PR. I would not want to separate this into another PR.

core/src/parser.rs
darrell-roberts: I've updated the comments for the two functions.

All the use cases are in the `core/data/tests/excluded_by_target_os/input.rs` snapshot test.

core/Cargo.toml
charlespierce: The workspace part makes sense, I was specifically asking about it moving from being under `dev-dependencies` to being under `dependencies`. Does `workspace` not work in `dev-dependencies`?

Cargo.toml
charlespierce: That's fair, and in this case I think it's small enough that it's not too much context switching when reviewing. In general though, I've found that getting timely reviews is easier when the changes are as small as possible. For example, this could be broken up by first having a PR that adds `log` and `flexi-logger` and converts all of the existing `println!` / `eprintln!` calls, then a separate PR stacked on top of that which adds the new behavior along side new uses of `debug!`.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Version 1.10.0-beta
 
-## 1.10.0-beta.7
```
charlespierce, line 3: nit: I think we should keep the existing changelog entry for `beta.7` and then add a new one. Also the updating the changelog is currently part of the release process, so not something that needs to be updated in a specific feature PR.

cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "typeshare-cli"
-version = "1.10.0-beta.7"
+version = "1.10.0-beta.8"
```
charlespierce, line 3: Similar to the changelog, the version bumps are part of the release process as well, so we should leave these out of specific feature PRs.

core/Cargo.toml
```diff
@@ -1,22 +1,24 @@
 [package]
 name = "typeshare-core"
-version = "1.10.0-beta.7"
+version = "1.10.0-beta.8"
```
charlespierce, line 3: Ditto here on the version changes.

core/src/parser.rs
```diff
@@ -618,54 +636,125 @@ fn is_redacted(attrs: &[syn::Attribute]) -> bool {
...
-                target_os_from_meta_list(meta_list)
+pub(crate) fn target_os_reject(attr: &Attribute, target_os: &str) -> bool {
+    if log_enabled!(log::Level::Debug) {
+        debug!(
+            "\tchecking attribute {} for {target_os} reject",
+            attr.into_token_stream()
+        );
+    }
+
+    let reject = get_meta_items(attr, "cfg").any(|meta| match &meta {
```
charlespierce, line 654: Does `get_meta_items` check recursively through nested lists? Specifically I'm wondering if this would work for the case of:

```
#[cfg(all(feature = "my-feature", not(target_os = "windows")))]
```

To reject Windows as a target OS.

core/src/parser.rs
```diff
@@ -618,54 +636,125 @@ fn is_redacted(attrs: &[syn::Attribute]) -> bool {
...
 }
 
-/// Check if we have a `target_os` cfg that dooes not match command line
-/// argument `--target-os`.
+/// Check if this attribute rejects our optional `target_os` arguments.
+///
+/// Examples:
+/// - `#[cfg(not(target_os = "windows"))]`
+/// - `#[cfg(not(any(target_os = "windows", target_os = "macos", feature = "my-feature")))]`
+/// - `#[cfg(not(all(target_os = "windows", feature = "my-feature")))]`
```
charlespierce, line 644: This config feels a little tricky—We are essentially ignoring the other configurations that aren't `target_os`, however `not(all(target_os = "windows", feature = "my-feature"))` will evaluate to `true` even on Windows if the feature flag isn't set. So in that case, Typeshare would be ignoring the attached declaration and not generating types, even though the compiler would actually be including the declaration.

Where we're making assumptions about how to parse configs like this, it feels like we should update the docs in the `docs/` directory to clearly indicate how the option behaves so that users aren't surprised by things not working the way they expect.

charlespierce (changes requested):

CHANGELOG.md
darrell-roberts: reverted

cli/Cargo.toml
darrell-roberts: reverted

core/Cargo.toml
darrell-roberts: reverted

core/src/parser.rs
darrell-roberts: I initially was going to avoid that, however I reworked it to handle recursive definitions now.

core/src/target_os_check.rs
```diff
@@ -0,0 +1,303 @@
...
+                }
+                Meta::List(meta_list) => {
+                    let nested_meta_list = meta_list
+                        .parse_args_with(Punctuated::<Meta, Token![,]>::parse_terminated)
+                        .inspect_err(|err| {
+                            error!("Failed to parse nested meta while traversing target_os candidates: {err}");
+                        })
+                        .ok()?;
+                    debug!("\texpanding with {} meta", nested_meta_list.len());
+                    self.meta.extend(nested_meta_list);
```
charlespierce, line 57: A few questions here to make sure I'm following the recursive parsing correctly:

- Since we're using a queue of `Meta` nodes, it looks like we're effectively doing a BFS through all of the attributes, is that correct?
- Is the `self.scope` used to track when we're inside of a `not()` block, so all of the equating needs to be inverted?

If I'm understanding correctly, does this handle the case where the `not` is a portion of a list? E.g.:

```
#[cfg(all(not(feature = "my-feature"), target_os = "android"))]
```

It appears it would first recurse into the `all`, then hit the `not` and toggle the `scope` to `Reject`, and then hit the `target_os` (which is actually _outside_ of the `not`) but return it with a scope of `Reject`.

FWIW I think it's reasonable not to support the infinitely nested configs and instead handle a subset of them—as long as we clearly document what is and isn't supported with the flags. We definitely don't need to let the perfect be the enemy of the good here, but I also want to make sure we're not creating extra work for ourselves as maintainers going forward if we add a feature that only works in specific cases but don't indicate what those cases are.

darrell-roberts: Updated scope handling and added this unit test.

core/Cargo.toml
darrell-roberts: Updated

core/src/target_os_check.rs
```diff
@@ -0,0 +1,329 @@
...
+        let test_struct: ItemStruct = parse_quote! {
+            #[cfg(any(target_os = "ios", feature = "test"))]
+            pub struct Excluded;
+        };
+
+        assert!(!accept_target_os(
+            &test_struct.attrs,
+            &["macos".into(), "android".into()]
+        ));
+    }
```
charlespierce, line 304: For our own benefit in the future, as well as the benefit of new users of this feature, is it too much of a burden to write a section of the docs about how we process the `target_os` directives? This test is an example that feels, on the surface, incorrect—if you compile the code with `--features test`, even on `macos`, the struct `Excluded` will actually be included. So whether the `target_os` matters or not depends on the feature set.

I absolutely believe that trying to fully recreate the `cfg` parser is beyond the scope of this feature, however since we are effectively ignoring any other directives beyond the `target_os` ones, we should make that clear so when we run into a question about that in the future we have a clear path to understanding why the behavior is the way it is.

core/src/target_os_check.rs
```diff
@@ -0,0 +1,329 @@
...
+        let test_struct: ItemStruct = parse_quote! {
+            #[cfg(not(any(target_os = "wasm32", target_os = "ios")))]
+            pub struct NestedNotTarget;
+        };
+
+        assert!(!accept_target_os(
+            &test_struct.attrs,
+            &["ios".into(), "android".into()]
+        ));
+    }
```
charlespierce, line 230: Similarly with this test, if we're running TypeShare with multiple `--target_os` settings, my intuitive thought is that we'll be using the types in _each_ of the output OSes. In that case, running the build for `android` would pass this check, because it is neither `wasm32` nor `ios`. However, this test indicates that it's a rejection, because of the overlap of having `--target_os ios,android`.

I don't have enough of a grasp of the underlying use-case for multiple `target_os` to make an argument for one approach or the other, however given that it's somewhat confusing, it would be beneficial to clearly document the logic that we use to determine whether a type is included or not.

core/src/target_os_check.rs
```diff
@@ -0,0 +1,329 @@
...
+
+impl Iterator for TargetOsIterator {
+    type Item = (TargetScope, String);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some((mut scope, meta)) = self.meta.pop() {
+            if meta.path().is_ident("not") {
+                debug!("encountered not");
+                scope = TargetScope::Reject
+            }
```
charlespierce, line 39: Should this take into account the existing scope? I.e., if we're already in `TargetScope::Reject` for this node, would `not` turn that into `TargetScope::Accept`? It's super edge-case since I don't know that someone would realistically do this, but how do we handle `#[cfg(not(not(target_os = "ios")))]`?

docs/src/usage/configuration.md
```diff
@@ -7,16 +7,22 @@ The behaviour of Typeshare can be customized by either passing options on the co
...
-    Specify a prefix that will be prepended to type names when generating types in Swift. 
+    Specify a prefix that will be prepended to type names when generating types in Swift.
 
 - `-M`, `--module-name`
-    Specify the name of the Kotlin module for generated Kotlin source code. 
+    Specify the name of the Kotlin module for generated Kotlin source code.
+
+- `-t`, `--target-os`
+    Optional comma separated list of target os targets. Types that are restricted via `#[cfg(target_os = <target>]`
+    that do not match the argument list will be filtered out.
```
charlespierce, line 22: Really appreciate having the flag documented here ❤️

charlespierce (changes requested):
There's been a lot of iteration on this PR, and I know the delay is frustrating. I definitely don't think we need to cover every possible edge case, however I do think it would be valuable—even if only for ourselves as future maintainers—to have a clear documentation of how the new flag works and why. That way, if we lose context but have to come back and re-evaluate, we can be more confident we're considering all of the cases that were originally considered.

core/src/target_os_check.rs
darrell-roberts: Sure. I added `target_os.md` to the typeshare book to describe how the feature works. I'm somewhat exploring what other use cases others may require. I had an initial use case, however after exposing others to this feature, additional uses cases came up which evolved into this PR.

core/src/target_os_check.rs
darrell-roberts: 🤔

charlespierce (approved):
LGTM! Thanks for going through all the iteration and for adding the docs 🎉

PR #189, opened 2024-08-05
Explicitly format variant types in Go

hculea:
## Background

https://github.com/1Password/typeshare/pull/183 introduced an edge-case regression in the handling of custom-typed algebraic enums that require uppercase acronyms handling in Go.

Specifically, the check:
```
custom_structs.contains(&variant_type.as_str())
```
would now fail, since it performs a case-sensitive check in the custom_structs hashmap. While this still produces working code, it does not handle the custom structs as expected. This PR fixes that.

## Effect

The bug consists of code that was previously generated as:
```
func (i ItemFieldDetails) OTP() *OtpFieldDetails {
	res, _ := i.content.(*OtpFieldDetails)
	return res
}
``` 

and would now be generated as:
```
func (i ItemFieldDetails) OTP() OTPFieldDetails {
	res, _ := i.content.(*OTPFieldDetails)
	return *res
}
```

instead of:
```
func (i ItemFieldDetails) OTP() *OTPFieldDetails {
	res, _ := i.content.(*OTPFieldDetails)
	return res
}
```

## Solution

The lookup in the hashmap has been made to leverage the default formatted struct name. However, for generating the Go code, the variant name is formatted with the provided uppercase acronyms.

### DIFF
--- a/core/src/language/go.rs
+++ b/core/src/language/go.rs
@@ -278,7 +278,6 @@ impl Go {
                     ));
 
                     if let Some(variant_type) = variant_type {
-                        let variant_type = self.acronyms_to_uppercase(&variant_type);
                         let (variant_pointer, variant_deref, variant_ref) =
                             match (v, custom_structs.contains(&variant_type.as_str())) {
                                 (RustEnumVariant::AnonymousStruct { .. }, ..) | (.., true) => {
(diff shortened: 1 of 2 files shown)


seanaye (approved):
This change is very straightforward, I'm comfortable merging provided all the tests pass

PR #190, opened 2024-08-06
Kotlin: hide value of redacted inline value classes

jonduran3000:
This pull request makes the `value` of any Kotlin inline value class, that is marked with `redacted`, private. It also provides an `unwrap()` function used for explicitly unwrapping the value. 

#### Input:

```rust
#[typeshare(serialized_as = "String", kotlin = "JvmInline", redacted)
struct MyType(String)
```

#### Output:

```kotlin
@Serializable
@JvmInline
value class MyType(private val value: String) {
  fun unwrap() = value
  
  override fun toString(): String = "***"
}
```


core/src/language/kotlin.rs
```diff
@@ -124,26 +124,46 @@ impl Language for Kotlin {
             writeln!(w, "@JvmInline")?;
             writeln!(w, "value class {}{}(", self.prefix, ty.id.renamed)?;
 
-            self.write_element(
-                w,
-                &RustField {
-                    id: Id {
-                        original: String::from("value"),
-                        renamed: String::from("value"),
+            if ty.is_redacted {
```
seanaye, line 127: could we express this as a match instead of if/else

core/src/language/kotlin.rs
```diff
@@ -426,6 +446,36 @@ impl Kotlin {
         )
     }
 
+    fn write_private_element(
```
seanaye, line 449: this function only differs from `write_element` by a single line. Instead of duplicating the function could we instead add a new argument to write_element. e.g.

```rust
enum Privacy {
    Public,
    Private
}

fn write_element(
        &mut self,
        w: &mut dyn Write,
        f: &RustField,
        generic_types: &[String],
        requires_serial_name: bool,
        privacy: Privacy
    ) -> std::io::Result<()> {
        ...
        write!(
            w,
            match privacy {
                Privacy::Public => "\tval {}: {}{}",
                Privacy::Private => "\tprivate val {}: {}{}"
            },
            ...
}
```

seanaye (approved):

PR #194, opened 2024-08-27
Optional fields default to none

amandayu1:


hculea:
To the repo maintainers: this can be closed, this is only an improvement that will be included as part of https://github.com/1Password/typeshare/pull/169

savannidgerinel:
@amandayu1 It does look like #169 will include these changes. Do you object to me closing this PR as @hculea suggested?

hculea:
@savannidgerinel - for more context around this - this is not an open-source contribution, as Amanda was an intern with the Programmatic Interfaces team in the summer term. This was part of the team's sprint, and this work was eventually included in the other PR that I referenced.

Also, since her internship ended, I believe Amanda no longer has access on the fork that contains this PR's source branch (which is under my namespace), so I think this should be safe to close 😄

PR #198, opened 2024-09-18
Avoid pointer redirection for Optional slices in Go

hculea:
In Go, optional values are represented by pointers. This makes sense for most types. For example, using a "*bool" pointer allows for the type to express three values: `nil`, `true` or `false`.

In the 1Password SDKs, we expose the types generated by typeshare (e.g. see [the types for Go](https://github.com/1Password/onepassword-sdk-go/blob/main/types.go)) to be user facing. 

Typeshare currently uses a pointer redirection to mark the Go types as optional. While most of the values are translated correctly, using a pointer to a slice (i.e. the translation of `Vec<T>` currently is `[]*string`) is not usual in Go, and it leads to more complex code for the users of the SDK.

This is because, in Go, slices are already reference types. That is, passing a slice around passes a reference to the underlying data.

However, an edge case can occur:
```
type A struct {
    Slice []string `json:",omitempty"`
}

type B struct {
    Slice *[]string `json:",omitempty"`
}
```

For type `A`, both `Slice: nil` and `Slice: []string{}` have the same JSON serialisation (`Slice` is omitted).

For type `B` `Slice: nil` and `Slice: []string{}` both have a different JSON serialisation. In the first case, `Slice` is omitted. In the second case the field has the value `[]`.

This, however, is rarely applicable in practice, and having this feature does not justify exposing an unintuitive user interface. Therefore, this PR is adding a configuration option disabling that for Go.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -51,6 +51,7 @@ pub struct GoParams {
     pub package: String,
     pub type_mappings: HashMap<String, String>,
     pub uppercase_acronyms: Vec<String>,
+    pub no_pointer_slice: bool,
 }
 
 /// The paramters that are used to configure the behaviour of typeshare
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -31,11 +31,11 @@ pub struct ScalaParams {
 #[serde(default)]
 pub struct SwiftParams {
     pub prefix: String,
-    pub type_mappings: HashMap<String, String>,
     pub default_decorators: Vec<String>,
     pub default_generic_constraints: Vec<String>,
     /// The contraints to apply to `CodableVoid`.
     pub codablevoid_constraints: Vec<String>,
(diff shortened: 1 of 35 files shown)


seanaye (commented):
Could we add a unit test to assert this setting works as expected in both cases

cli/src/config.rs
```diff
@@ -50,6 +50,7 @@ pub struct TypeScriptParams {
 pub struct GoParams {
     pub package: String,
     pub uppercase_acronyms: Vec<String>,
+    pub no_pointer_slice: bool,
```
seanaye, line 53: The description of the problem in the PR was very good. Could we copy/paste some of that prose here as a doc comment

hculea: Done!

seanaye (approved):

PR #202, opened 2024-10-09
Remove atty by migrating to clap 4

Lucretiel:
This PR removes the dependency on `atty`, which has been deprecated, by migrating to clap 4. In the course of this migration, we replaced the builder-style command line arguments with a simpler derive-style. The bulk of this PR is dedicated to ensuring the migration has parity with the original functionality.


savannidgerinel (approved):
We did an interactive review and all of the changes look good.

srriddle (approved):

PR #205, opened 2024-11-12
Only overwrite Codable.swift if it has changed.

kevinbhayes:
typeshare was always writing Codable.swift when generating multiple files. The issue is that the updated modification date on Codable.swift would trigger Xcode to rebuild the module, even if there were no changes.

This PR now checks the contents of Codable.swift and only overwrites it if the contents need changing.

### DIFF
--- a/core/src/language/swift.rs
+++ b/core/src/language/swift.rs
@@ -758,12 +758,11 @@ impl Swift {
     fn write_codable_file(&self, output_folder: &str) -> std::io::Result<()> {
         let output_string = self.get_codable_contents();
         let output_path = Path::new(output_folder).join("Codable.swift");
-        match fs::read(output_path.clone()) {
-            Ok(buf) if buf == output_string.as_bytes() => {
-                // don't overwrite the same file, we can bail here
+
+        if let Ok(buf) = fs::read(&output_path) {
+            if buf == output_string.as_bytes() {
                 return Ok(());
             }
-            _ => {}
         }
 
         let mut w = fs::File::create(output_path)?;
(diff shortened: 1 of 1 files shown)


core/src/language/swift.rs
```diff
@@ -756,18 +756,21 @@ impl Swift {
...
-    }
+        let output_string = self.get_codable_contents();
+        let output_path = Path::new(output_folder).join("Codable.swift");
+        match fs::read(output_path.clone()) {
+            Ok(buf) if buf == output_string.as_bytes() => {
+                // don't overwrite the same file, we can bail here
+                return Ok(());
+            }
+            _ => {}
+        }
```
darrell-roberts, line 767: Avoid cloning and update for single pattern.

```rust
        if let Ok(buf) = fs::read(&output_path) {
            if buf == output_string.as_bytes() {
                return Ok(());
            }
        }
```

core/src/language/swift.rs
```diff
@@ -778,7 +781,12 @@ impl Swift {
             decs.push(CODABLE);
         }
 
-        writeln!(w, "public struct CodableVoid: {} {{}}", decs.join(", "))
+        format!("\n/// () isn't codable, so we use this instead to represent Rust's unit type\npublic struct CodableVoid: {} {{}}\n", decs.join(", "))
+    }
```
darrell-roberts, line 785: These two snapshot tests will need to be updated for this new output:

    can_generate_generic_struct::swift
    can_handle_unit_type::swift

You can run:

```bash
UPDATE_EXPECT=1 cargo t --all-features --test snapshot_tests can_generate_generic_struct
UPDATE_EXPECT=1 cargo t --all-features --test snapshot_tests can_handle_unit_type
```

darrell-roberts (approved):
LGTM

seanaye (approved):

Issue #201, opened 2024-10-03
#[serde(rename = "_")]


HHogg:
```rust
#[derive(Serialize)]
#[serde(rename = "SomethingFoo")]
#[typeshare]
pub enum Foo {
  A  
}

#[derive(Serialize)]
#[typeshare]
#[serde(tag = "type", content = "value")]
pub enum Parent {
  B(Foo),
}
```

Outputs

```ts
export enum SomethingFoo = { A =  'A' }

export type Parent = { type: 'B', value: Foo  } // No Foo. Only SomethingFoo
```

darrell-roberts:
I have a WIP solution for this: https://github.com/1Password/typeshare/pull/210. 

Are you using the multiple file output option? If your outputting to a single file this approach works, however in the case of multiple files it assumes, at the moment, your renamed type and reference are both in the same crate scope. 

Would you be able to try this out and let me know if it covers your use cases? I created this [snapshot](https://github.com/1Password/typeshare/blob/00a44c619b539617424ec254e094bebbaa68b8de/core/data/tests/serde_rename_references/input.rs) test which adds additional reference use cases.

PR #210, opened 2024-11-17
Approach to handle serde(rename) references.

darrell-roberts:
I added a post parsing reconcile step that updates reference types that refer to a type that has been renamed via `serde(rename)`. 

https://github.com/1Password/typeshare/issues/201


seanaye (approved):
LGTM!

PR #216, opened 2024-11-23
Use parallel walker instead of rayon

darrell-roberts:
The ignore walker we use to walk the list of directories recursively already has built in capabilities for parallel walking. This PR uses that instead of the existing rayon solution.

This PR also refactors the cli code so the main function is broken up into separate functions by separate concerns.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -2,48 +2,47 @@
 use anyhow::Context;
 use crossbeam::channel::bounded;
 use ignore::{DirEntry, WalkBuilder, WalkState};
-use log::error;
 use std::{
     collections::{BTreeMap, HashMap},
-    mem,
-    ops::Not,
-    path::PathBuf,
-    thread,
+    mem, thread,
 };
 use typeshare_core::{
     context::{ParseContext, ParseFileContext},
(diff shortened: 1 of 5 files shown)


cli/src/parse.rs
```diff
@@ -67,67 +63,84 @@ fn output_file_name(language_type: SupportedLanguage, crate_name: &CrateName) ->
...
+            let parsed_data = result?;
+            let crate_name = parsed_data.crate_name.clone();
+            // Append each yielded parsed data by its respective crate.
+            *crate_parsed_data.entry(crate_name).or_default() += parsed_data;
+        }
+
+        Ok(crate_parsed_data)
+    });
+
+    walker_builder.build_parallel().run(|| {
```
seanaye, line 122: does this outer closure run exactly 1 time? If no, then I think it would be better to create the boxed closure outside. If yes then disregard

seanaye (approved):

cli/src/parse.rs
darrell-roberts: The outer closure is called on each dir entry yielded. The inner for each thread.

https://docs.rs/ignore/latest/ignore/struct.WalkParallel.html#method.run

Similar to parallel walker example: https://github.com/BurntSushi/ripgrep/blob/master/crates/ignore/examples/walk.rs#L28.

Issue #44, opened 2022-12-24
Support for constants


d4h0:
Hi,

I tried the following:

```rust
#[typeshare::typeshare]
const SHARED_VAR: &str = "value";
```

...hoping that I could use this to define some shared constants in one place, but that doesn't seem to be supported.

There also wasn't an error message, which isn't ideal, I think.

InquisitivePenguin:
I think this could be a useful feature! However, we would probably have to limit the scope of this change to numeric and string constants for now.

d4h0:
@InquisitivePenguin: That seem reasonable. Booleans probably also could be supported easily.

dwosk:
@InquisitivePenguin any update on your PR? Support for Rust `&str` constants would be really useful for us, thanks!

InquisitivePenguin:
@dwosk Sorry about the delay - support for this should be ready soon.

PR #218, opened 2024-12-12
Add const support for integer types

edif2008:
This PR contains the functionality to parse Rust integer constants and generate them into the following languages:
- Go
- Typescript
- Python

The foundational pieces can easily be extended to other constant types (e.g. Boolean, Float and String). However, additional guidance and support might be needed to ensure it works as expected for the other types as well.

This is heavily inspired from #57. It was adapted to the latest version from `main` and added the code generation snippets for Go, Typescript and Python.

Relates: #44

### DIFF
--- a/core/data/tests/can_generate_const/output.go
+++ b/core/data/tests/can_generate_const/output.go
@@ -2,4 +2,4 @@ package proto
 
 import "encoding/json"
 
-const MyVar uint32 = 12;
+const MyVar uint32 = 12
--- a/core/data/tests/can_generate_const/output.go
+++ b/core/data/tests/can_generate_const/output.go
@@ -1,0 +1,5 @@
+package proto
+
+import "encoding/json"
+
+const MyVar uint32 = 12;
--- a/core/src/parser.rs
+++ b/core/src/parser.rs
@@ -3,9 +3,9 @@ use crate::{
     language::{CrateName, SupportedLanguage},
     rename::RenameExt,
     rust_types::{
-        DecoratorMap, FieldDecorator, Id, RustEnum, RustEnumShared, RustEnumVariant,
-        RustEnumVariantShared, RustField, RustItem, RustStruct, RustType, RustTypeAlias,
-        RustTypeParseError,
+        DecoratorMap, FieldDecorator, Id, RustConst, RustConstExpr, RustEnum, RustEnumShared,
+        RustEnumVariant, RustEnumVariantShared, RustField, RustItem, RustStruct, RustType,
+        RustTypeAlias, RustTypeParseError, SpecialRustType,
     },
     target_os_check::accept_target_os,
     visitors::{ImportedType, TypeShareVisitor},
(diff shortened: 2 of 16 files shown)


core/src/parser.rs
```diff
@@ -498,6 +510,66 @@ pub(crate) fn parse_type_alias(t: &ItemType) -> Result<RustItem, ParseError> {
...
+
+fn parse_const_expr(e: &Expr) -> Result<RustConstExpr, ParseError> {
+    struct ExprLitVisitor(pub Option<Result<RustConstExpr, ParseError>>);
+    impl Visit<'_> for ExprLitVisitor {
+        fn visit_expr_lit(&mut self, el: &ExprLit) {
+            // if self.0.is_some() {
+            //     // should we throw an error instead of silently ignoring a second literal?
+            //     // or would this create false positives?
+            //     return;
+            // }
```
Marton6, line 550: Do we want to keep this code or remove it completely? I don't think there's a good reason to keep around commented code.

edif2008: Shoot I commented it while debugging, but it should stay there. Nice catch. 🤦🏻

core/data/tests/can_generate_const/output.go
```diff
@@ -0,0 +1,5 @@
+package proto
+
+import "encoding/json"
+
+const MyVar uint32 = 12;
```
hculea, line 5: I would not expect a semicolumn here:
```suggestion
const MyVar uint32 = 12
```

core/src/parser.rs
```diff
@@ -498,6 +510,66 @@ pub(crate) fn parse_type_alias(t: &ItemType) -> Result<RustItem, ParseError> {
...
+        RustType::try_from(c.ty.as_ref())?
+    };
+
+    match &ty {
+        RustType::Special(SpecialRustType::HashMap(_, _))
+        | RustType::Special(SpecialRustType::Vec(_))
+        | RustType::Special(SpecialRustType::Option(_)) => {
+            return Err(ParseError::RustConstTypeInvalid);
+        }
+        RustType::Special(_s) => (),
```
hculea, line 531: Pro-tip: simply adding this line
```
        RustType::Simple { .. } => (),
```
here would give us support already for all the type aliases to int, which could be very useful if we want to not allow all int values to be considered valid.

This requires no additional validation, as we're still validating, in `parse_const_value`, that the value attributed is an int. Therefore, the line below would allow every type that can be attributed an int.

core/src/parser.rs
```diff
@@ -74,6 +74,10 @@ pub enum ParseError {
     SerdeTagRequired { enum_ident: String },
     #[error("serde content attribute needs to be specified for algebraic enum {enum_ident}. e.g. #[serde(tag = \"type\", content = \"content\")]")]
     SerdeContentRequired { enum_ident: String },
+    #[error("the expression assigned to this constant variable is not a numeric literal")]
+    RustConstExprInvalid,
+    #[error("you cannot use typeshare on a constant that is not a numeric literal")]
+    RustConstTypeInvalid,
```
hculea, line 80: Can we document these limitations in an issue? We want to make it as easy as possible for a contributor to pick this up, if they want to expand typeshare.

hculea: It should actually be two issues: 
* one about adding constants other than int types
* one about adding constants for Swift, Scala and Kotlin

darrell-roberts (approved):

core/src/parser.rs
edif2008: Issues created: #228, #229

PR #230, opened 2025-02-18
Add support for byte translation

MOmarMiraj:
This PR adds support for translating `Vec<u8>` to the appropriate byte translation for Go, Python, and Typescript. 

To maintain backwards compatibility, decided to opt against adding a `Bytes` in `SpecialRustType` as there could be some devs translating `Vec<u8>` into a non byte translation and will be a breaking change. 

There was two options I was considering when implementing this feature:
1) Adding it as a decorator so the user could do something like this:
```
#[typeshare]
#[serde(rename_all = "camelCase")]
pub struct Foo {
    #[typeshare(go,python,typescript(binary_data))]
    pub bytes: Vec<u8>,
}
```
This option would allow us to use the existing decorators functionality to parse each specific field to determine whether or not we can translate it to its appropriate byte translation. 

The issue with this implementation is it would cause a pretty big refactor as the `format_special_type` function does not have access to decorators and allowing the function to have access whether passing it as an argument or attaching it to each language struct as a field would be overly complex for achieving not the greatest result.

The second option is what this PR does, which is take adavantage of the `type_mappings` field which allows us to use the `typeshare.toml` to translate the `Vec<u8>` to its appropriate translation in each target language. This way is the least complex and is backwards compatible as the user would have to specify inside the `typeshare.toml` if they want the byte translation otherwise it would keep the current implementation intact.

I added some tests in Go, Python, and Typescript to test this out.

With typescript, as there is no native support for `Uint8Array` for JSON serialization/deserialization, we have to add these `reviver` and `replacer` functions to not lose context of the `Uint8Array` during serialization/deserialization and the same thing goes for the Python implementation.

### DIFF
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -1,21 +1,18 @@
 from __future__ import annotations
 
-from pydantic import BaseModel, field_validator
-from typing import Any
+from pydantic import BaseModel, field_serializer, field_validator
 
 
 class Foo(BaseModel):
     bytes: bytes
 
-    @field_validator('content', mode='before')
-    @classmethod
-    def cast_list_to_bytes(cls, value: Any) -> bytes:
-        if isinstance(value, list) and all(isinstance(i, int) for i in value):
-          return bytes(value)
-
-        raise ValueError("content must be a list of integers")
-
-    class Config:
-        json_encoders = {
-            bytes: lambda b: list(b),
-        }
+    
+    @field_serializer("content")
+    def serialize_data(self, value: bytes) -> list[int]:
+        return list(value)
+                
+    @field_validator("content", mode="before")
+    def deserialize_data(cls, value):
+        if isinstance(value, list): 
+            return bytes(value)
+        return value
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -1,24 +1,23 @@
 from __future__ import annotations
 
-from pydantic import BaseModel, ConfigDict, Field, field_serializer, field_validator
+from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSerializer
+from typing import Annotated
 
 
+def deserialize_data(value):
+    if isinstance(value, list):
+        if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+            return bytes(value)
+        raise ValueError("All elements must be integers in the range 0-255 (u8).")
+    elif isinstance(value, bytes):
+            return value
+    raise TypeError("Content must be a list of integers (0-255) or bytes.")
+            
+            
+def serialize_data(value: bytes) -> list[int]:
+    return list(value)
 class Foo(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
-    this_is_bits: bytes = Field(alias="thisIsBits")
+    this_is_bits: Annotated[bytes, BeforeValidator(deserialize_data), PlainSerializer(serialize_data)] = Field(alias="thisIsBits")
 
-    
-    @field_serializer("thisIsBits")
-    def serialize_data(self, value: bytes) -> list[int]:
-        return list(value)
-                
-    @field_validator("thisIsBits", mode="before")
-    def deserialize_data(cls, value):
-        if isinstance(value, list):
-            if all(isinstance(x, int) and 0 <= x <= 255 for x in value): 
-                return bytes(value) 
-            raise ValueError("All elements must be integers in the range 0-255 (u8).")
-        elif isinstance(value, bytes):
-            return value
-        raise TypeError("thisIsBits must be a list of integers (0-255) or bytes.")
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -4,7 +4,7 @@ from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSeriali
 from typing import Annotated
 
 
-def deserialize_data(value):
+def deserialize_binary_data(value):
     if isinstance(value, list):
         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
             return bytes(value)
@@ -14,10 +14,11 @@ def deserialize_data(value):
     raise TypeError("Content must be a list of integers (0-255) or bytes.")
             
             
-def serialize_data(value: bytes) -> list[int]:
+def serialize_binary_data(value: bytes) -> list[int]:
     return list(value)
+
 class Foo(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
-    this_is_bits: Annotated[bytes, BeforeValidator(deserialize_data), PlainSerializer(serialize_data)] = Field(alias="thisIsBits")
+    this_is_bits: Annotated[bytes, BeforeValidator(deserialize_binary_data), PlainSerializer(serialize_binary_data)] = Field(alias="thisIsBits")
 
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -4,18 +4,17 @@ from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSeriali
 from typing import Annotated
 
 
-def deserialize_binary_data(value):
-    if isinstance(value, list):
-        if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
-            return bytes(value)
-        raise ValueError("All elements must be integers in the range 0-255 (u8).")
-    elif isinstance(value, bytes):
-            return value
-    raise TypeError("Content must be a list of integers (0-255) or bytes.")
-            
-            
 def serialize_binary_data(value: bytes) -> list[int]:
-    return list(value)
+        return list(value)
+
+def deserialize_binary_data(value):
+     if isinstance(value, list):
+         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+             return bytes(value)
+         raise ValueError("All elements must be integers in the range 0-255 (u8).")
+     elif isinstance(value, bytes):
+             return value
+     raise TypeError("Content must be a list of integers (0-255) or bytes.")
 
 class Foo(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -1,18 +1,24 @@
 from __future__ import annotations
 
-from pydantic import BaseModel, field_serializer, field_validator
+from pydantic import BaseModel, ConfigDict, Field, field_serializer, field_validator
 
 
 class Foo(BaseModel):
-    bytes: bytes
+    model_config = ConfigDict(populate_by_name=True)
+
+    this_is_bits: bytes = Field(alias="thisIsBits")
 
     
-    @field_serializer("content")
+    @field_serializer("thisIsBits")
     def serialize_data(self, value: bytes) -> list[int]:
         return list(value)
                 
-    @field_validator("content", mode="before")
+    @field_validator("thisIsBits", mode="before")
     def deserialize_data(cls, value):
-        if isinstance(value, list): 
-            return bytes(value)
-        return value
+        if isinstance(value, list):
+            if all(isinstance(x, int) and 0 <= x <= 255 for x in value): 
+                return bytes(value) 
+            raise ValueError("All elements must be integers in the range 0-255 (u8).")
+        elif isinstance(value, bytes):
+            return value
+        raise TypeError("thisIsBits must be a list of integers (0-255) or bytes.")
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from pydantic import BaseModel, field_validator
+from typing import Any
+
+
+class Foo(BaseModel):
+    bytes: bytes
+
+    @field_validator('content', mode='before')
+    @classmethod
+    def cast_list_to_bytes(cls, value: Any) -> bytes:
+        if isinstance(value, list) and all(isinstance(i, int) for i in value):
+          return bytes(value)
+
+        raise ValueError("content must be a list of integers")
+
+    class Config:
+        json_encoders = {
+            bytes: lambda b: list(b),
+        }
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -1,15 +1,13 @@
 export interface Foo {
-	bytes: Uint8Array;
+	thisIsBits: Uint8Array;
 }
 
-// Reviver code  - required for JSON deserialization
 export function TypeshareReviver(key: string, value: unknown): unknown {
-    return Array.isArray(value) && value.every(Number.isFinite) 
+    return Array.isArray(value) && value.every(v => Number.isFinite(v) && v >= 0 && v <= 255)  
         ? new Uint8Array(value) 
         : value;
 }
-    
-// Replacer code - required for JSON serialization
+
 export function TypeshareReplacer(key: string, value: unknown): unknown {
     if (value instanceof Uint8Array) {
         return Array.from(value);
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -3,7 +3,7 @@ export interface Foo {
 }
 
 export function ReviverFunc(key: string, value: unknown): unknown {
-    return Array.isArray(value) && value.every(v => Number.isFinite(v) && v >= 0 && v <= 255)  
+    return Array.isArray(value) && value.every(v => Number.IsInteger(v) && v >= 0 && v <= 255)  
         ? new Uint8Array(value) 
         : value;
 }
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -3,7 +3,7 @@ export interface Foo {
 }
 
 export function ReviverFunc(key: string, value: unknown): unknown {
-    return Array.isArray(value) && value.every(v => Number.IsInteger(v) && v >= 0 && v <= 255)  
+    return Array.isArray(value) && value.every(v => Number.isInteger(v) && v >= 0 && v <= 255)  
         ? new Uint8Array(value) 
         : value;
 }
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -3,17 +3,14 @@ export interface Foo {
 }
 
 // Reviver code  - required for JSON deserialization
-function TypeshareReviver(key: string, value: unknown): unknown { 
-    return isNumberArray(value) ? new Uint8Array(value) : value; 
+export function TypeshareReviver(key: string, value: unknown): unknown {
+    return Array.isArray(value) && value.every(Number.isFinite) 
+        ? new Uint8Array(value) 
+        : value;
 }
-            
-function isNumberArray(value: unknown): value is number[] {
-    return Array.isArray(value) && value.every(item => typeof item === "number");
-}
-            
+    
 // Replacer code - required for JSON serialization
-            
-function TypeshareReplacer(key: string, value: unknown): unknown {
+export function TypeshareReplacer(key: string, value: unknown): unknown {
     if (value instanceof Uint8Array) {
         return Array.from(value);
     }
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -2,13 +2,13 @@ export interface Foo {
 	thisIsBits: Uint8Array;
 }
 
-export function TypeshareReviver(key: string, value: unknown): unknown {
+export function ReviverFunc(key: string, value: unknown): unknown {
     return Array.isArray(value) && value.every(v => Number.isFinite(v) && v >= 0 && v <= 255)  
         ? new Uint8Array(value) 
         : value;
 }
 
-export function TypeshareReplacer(key: string, value: unknown): unknown {
+export function ReplacerFunc(key: string, value: unknown): unknown {
     if (value instanceof Uint8Array) {
         return Array.from(value);
     }
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -3,9 +3,10 @@ export interface Foo {
 }
 
 export function ReviverFunc(key: string, value: unknown): unknown {
-    return Array.isArray(value) && value.every(v => Number.isInteger(v) && v >= 0 && v <= 255)  
-        ? new Uint8Array(value) 
-        : value;
+    if (Array.isArray(value) && value.every(v => Number.isInteger(v) && v >= 0 && v <= 255)) {
+                    return new Uint8Array(value);
+                }
+    return value;
 }
 
 export function ReplacerFunc(key: string, value: unknown): unknown {
--- a/core/data/tests/test_byte_translation/output.ts
+++ b/core/data/tests/test_byte_translation/output.ts
@@ -1,0 +1,21 @@
+export interface Foo {
+	bytes: Uint8Array;
+}
+
+// Reviver code  - required for JSON deserialization
+function TypeshareReviver(key: string, value: unknown): unknown { 
+    return isNumberArray(value) ? new Uint8Array(value) : value; 
+}
+            
+function isNumberArray(value: unknown): value is number[] {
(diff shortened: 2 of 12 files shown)


core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from pydantic import BaseModel, field_serializer, field_validator
+
+
+class Foo(BaseModel):
+    bytes: bytes
+
+    
+    @field_serializer("content")
```
hculea, line 10: This should be the name of the field that the serializer targets, i.e. `bytes` in this situation.

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from pydantic import BaseModel, field_serializer, field_validator
+
+
+class Foo(BaseModel):
+    bytes: bytes
```
hculea, line 7: Isn't bytes a reserved word? I don't think a field can be named as such.

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,18 @@
...
+
+    
+    @field_serializer("content")
+    def serialize_data(self, value: bytes) -> list[int]:
+        return list(value)
+                
+    @field_validator("content", mode="before")
```
hculea, line 14: Same comment as above for this - shouldn't be `"content"`.

core/data/tests/test_byte_translation/output.ts
```diff
@@ -0,0 +1,18 @@
+export interface Foo {
+	bytes: Uint8Array;
+}
+
+// Reviver code  - required for JSON deserialization
```
hculea, line 5: These comments (this and below) should not be required.

core/src/language/go.rs
```diff
@@ -119,7 +119,14 @@ impl Language for Go {
         generic_types: &[String],
     ) -> Result<String, RustTypeFormatError> {
         Ok(match special_ty {
-            SpecialRustType::Vec(rtype) => format!("[]{}", self.format_type(rtype, generic_types)?),
+            SpecialRustType::Vec(rtype) => {
+                // TODO: https://github.com/1Password/typeshare/issues/231
+                if let Some(conversion) = get_vec_u8_conversion(special_ty, self.type_map(), rtype)
+                {
+                    return Ok(conversion);
```
hculea, line 126: I don't think we should call this function for every `Vec` type, but instead validate that `rtype` is indeed a box of u8. We can also exclude this validation from the function's implementation.

hculea: This will even help remove the function all together I feel like, as you'd just have to return the type mapping.

core/src/language/python.rs
```diff
@@ -174,9 +175,16 @@ impl Language for Python {
...
-            SpecialRustType::Vec(rtype)
-            | SpecialRustType::Array(rtype, _)
-            | SpecialRustType::Slice(rtype) => {
+            SpecialRustType::Vec(rtype) => {
+                // TODO: https://github.com/1Password/typeshare/issues/231
+                if let Some(conversion) = get_vec_u8_conversion(special_ty, self.type_map(), rtype)
+                {
+                    return Ok(conversion);
+                }
+                self.add_import("typing".to_string(), "List".to_string());
```
hculea, line 184: Same comment as above.

core/src/language/python.rs
```diff
@@ -294,8 +303,31 @@ impl Language for Python {
...
         }
         writeln!(w)?;
-        self.add_import("pydantic".to_string(), "BaseModel".to_string());
-        Ok(())
+        rs.fields.iter().try_for_each(|field| {
+            let python_type = self
+                .format_type(&field.ty, &rs.generic_types)
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+            // Ideally, we should statically compare this to SpecialRustType::Bytes
+            // This would cause too much of a refactor
```
hculea, line 311: Have you opened a follow-up ticket for this?

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Its not a reserved word and can be used but its generally discouraged. Considering this is a test, this shouldn't matter.

core/src/parser.rs
```diff
@@ -80,6 +80,8 @@ pub enum ParseError {
     RustConstTypeInvalid,
     #[error("the serde flatten attribute is not currently supported")]
     SerdeFlattenNotAllowed,
+    #[error("the binary data decorator isn't allowed on types other than Vec<u8>")]
+    BinaryDataOnTypesOtherThanBytesNotAllowed,
```
hculea, line 84: Are we still using this?

core/src/language/typescript.rs
```diff
@@ -153,7 +158,32 @@ impl Language for TypeScript {
             .iter()
             .try_for_each(|f| self.write_field(w, f, rs.generic_types.as_slice()))?;
 
-        writeln!(w, "}}\n")
+        writeln!(w, "}}\n")?;
+        rs.fields.iter().try_for_each(|field| {
+            let typescript_type = self
+                .format_type(&field.ty, &rs.generic_types)
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+            if typescript_type == "Uint8Array" {
```
hculea, line 166: I dislike this approach of comparing strings. Since these functions should only be outputted once regardless of how many times we encounter a byte array, can we make this a global field on the global `Typescript` struct and set it to true in the `format_special_type` function?

core/src/language/python.rs
hculea: A better way to do this is to format it within the type and define the transformation functions globally, to be able to do this:
```
def validator_func(value: str) -> int:
    // converts from a str to int when deserializing

def age_serializer(value: int) -> str:
    // converts from an int to a str when serializing

class Example(BaseModel):
    age: Annotated[int, BeforeValidator(coerce_age), PlainSerializer(age_serializer)]
```

I have not tested this, but looks like a cleaner option, such that we don't have to duplicate the serializer/deserializer for every field that contains binary data

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Done!

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Done!

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,18 @@
...
+    def deserialize_data(cls, value):
+        if isinstance(value, list):
```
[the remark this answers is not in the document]
MOmarMiraj: Done!

core/data/tests/test_byte_translation/output.ts
```diff
@@ -0,0 +1,18 @@
...
+export function TypeshareReviver(key: string, value: unknown): unknown {
+    return Array.isArray(value) && value.every(Number.isFinite)
```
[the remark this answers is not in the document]
MOmarMiraj: Added the check!

core/src/language/go.rs
MOmarMiraj: Right, fixed that!

core/src/language/python.rs
MOmarMiraj: Yeah forgot to add the TODO, Added!

core/src/parser.rs
MOmarMiraj: Nope deleted, thanks for the catch.

core/src/language/typescript.rs
MOmarMiraj: Didn't like it either and changed!

core/src/language/mod.rs
```diff
@@ -492,6 +492,17 @@ fn used_imports<'a, 'b: 'a>(
...
+/// the typeshare.toml
+fn get_vec_u8_conversion(
+    special_type: &SpecialRustType,
+    type_mappings: &HashMap<String, String>,
+    inner_type: &RustType,
+) -> Option<String> {
+    type_mappings
+        .get(&format!("{}<{}>", special_type.id(), inner_type.id()))
+        .map(ToString::to_string)
+}
```
hculea, line 505: Is this function even needed? This can simply be `type_mappings
        .get("Vec<u8>")
        .map(ToString::to_string)`, I don't think this warrants its own function.

core/src/language/python.rs
MOmarMiraj: I have tested this on a POC and seems to work nice! changed it.

core/src/language/mod.rs
MOmarMiraj: Changed!

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,23 @@
...
+            return bytes(value)
+        raise ValueError("All elements must be integers in the range 0-255 (u8).")
+    elif isinstance(value, bytes):
+            return value
+    raise TypeError("Content must be a list of integers (0-255) or bytes.")
+            
+            
+def serialize_data(value: bytes) -> list[int]:
+    return list(value)
+class Foo(BaseModel):
```
hculea, line 19: could we have a new line between these?

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSerializer
+from typing import Annotated
+
+
+def deserialize_data(value):
```
hculea, line 7: let's call this `deserialize_binary_data`

core/data/tests/test_byte_translation/output.py
```diff
@@ -0,0 +1,23 @@
...
+    if isinstance(value, list):
+        if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+            return bytes(value)
+        raise ValueError("All elements must be integers in the range 0-255 (u8).")
+    elif isinstance(value, bytes):
+            return value
+    raise TypeError("Content must be a list of integers (0-255) or bytes.")
+            
+            
+def serialize_data(value: bytes) -> list[int]:
```
hculea, line 17: same for this - `serialize_binary_data`.

core/src/language/go.rs
```diff
@@ -119,7 +119,17 @@ impl Language for Go {
...
+                // TODO: https://github.com/1Password/typeshare/issues/231
+                if rtype.contains_type(SpecialRustType::U8.id()) {
+                    if let Some(conversion) =
+                        self.type_map().get("Vec<u8>").map(ToString::to_string)
+                    {
+                        return Ok(conversion);
+                    }
+                }
+                format!("[]{}", self.format_type(rtype, generic_types)?)
+            }
```
hculea, line 132: This is a comment that goes for all the implementations:

The fact that we're mapping these types to the types that the user provides is not specific to Vec<u8>.

This should happen at the beginning of the function, for all the types that have type mappings.

Something like:
```
fn format_special_type(...) ->
  if let Some(mapped) = self.type_map().get(special_type) 
        {
            mapped.into()
        }
  <actual-translation-logic>
```

What should be in this branch, in the translation logic, however, is the rest of the logic, which is:
* nothing for Go
* setting the `is_bytes` bool for Typescript
* (not sure what that is in Python, still parsing through the code).

hculea: To sum it up, in TS, this would look like this:
```
fn format_special_type(...) ->
  if let Some(mapped) = self.type_map().get(special_ty) 
        {
            mapped.into()
        }
  match special_ty {
            SpecialRustType::Vec(rtype) => {
                // if rtype is u8 and there is a type mapping for it which is exactly Uint8Array, output the reviver and the replacer
            }
```

core/src/language/python.rs
```diff
@@ -390,22 +437,22 @@ impl Python {
...
-        match (not_optional_but_default, is_aliased) {
-            (true, true) => {
-                self.add_import("typing".to_string(), "Optional".to_string());
-                self.add_import("pydantic".to_string(), "Field".to_string());
+        // the type is checked to ensure its printed on the actual field that is being byte translated and not all
+        let is_bytes = matches!(&field.ty, RustType::Special(SpecialRustType::Vec(boxed_type)) if matches!(**boxed_type, RustType::Special(SpecialRustType::U8)))
+            && self.is_bytes;
+        // Adds all the required imports needed based off whether its optional ,aliased, or needs a byte translation
+        self.add_common_imports(is_optional, is_bytes, is_aliased);
+        match (not_optional_but_default, is_aliased, is_bytes) {
```
hculea, line 445: I don't think this is a very scalable way to do this. 

In the future, we may want to add custom translations and serializers/deserializers for a bunch of other types (dates, durations), this will quickly go unscalable if we explicitly check if this is of a specific type statically.

Instead, what I recommend is that we maintain a HashMap from the custom type to its serialization/deserialization functions and that this match below takes care of everything else.

For example:
```
let map = HashMap::from([("bytes", ("serialize_bytes", "deserialize_bytes"))]);
```

The goal here is that, when we add another type that requires custom serialization/deserialization, we can simply add it to this hashmap, make sure that these functions are implemented (as you now do at the top of the file for the bytes ones) and everything else will be taken care of.

core/src/language/typescript.rs
```diff
@@ -153,7 +166,27 @@ impl Language for TypeScript {
             .iter()
             .try_for_each(|f| self.write_field(w, f, rs.generic_types.as_slice()))?;
 
-        writeln!(w, "}}\n")
+        writeln!(w, "}}\n")?;
+        if self.is_bytes {
+            self.is_bytes = false;
+            writeln!(
```
hculea, line 172: This logic is pretty ugly here, can we include these serializers/deserializers in their own block, e.g. via a `end_file` function as this PR implemented? https://github.com/1Password/typeshare/pull/85/files

core/data/tests/test_byte_translation/output.ts
```diff
@@ -0,0 +1,16 @@
...
+	thisIsBits: Uint8Array;
+}
+
+export function TypeshareReviver(key: string, value: unknown): unknown {
+    return Array.isArray(value) && value.every(v => Number.isFinite(v) && v >= 0 && v <= 255)  
+        ? new Uint8Array(value) 
+        : value;
+}
+
+export function TypeshareReplacer(key: string, value: unknown): unknown {
```
hculea, line 11: I don't think using `Typeshare` in the name of these structs is good. Typeshare is an internal tool, we don't want the code written leveraging its types to be "leaking" implementation details like that.

Can we call this simply `ReviverFunc` and `ReplacerFunc`?

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Done!

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Fixed!

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Fixed!

core/src/language/go.rs
MOmarMiraj: Fixed this to be more scaleable so in the future, if there is more if these the logic doesn't have to be changed to much!

core/src/language/python.rs
```diff
@@ -390,22 +437,22 @@ impl Python {
             .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
         let python_field_name = python_property_aware_rename(&field.id.original);
         let is_aliased = python_field_name != field.id.renamed;
-        match (not_optional_but_default, is_aliased) {
-            (true, true) => {
-                self.add_import("typing".to_string(), "Optional".to_string());
-                self.add_import("pydantic".to_string(), "Field".to_string());
+        // the type is checked to ensure its printed on the actual field that is being byte translated and not all
+        let is_bytes = matches!(&field.ty, RustType::Special(SpecialRustType::Vec(boxed_type)) if matches!(**boxed_type, RustType::Special(SpecialRustType::U8)))
+            && self.is_bytes;
```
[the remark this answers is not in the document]
MOmarMiraj: changed this logic to match on the output of `format_type` which is `python_type`

core/src/language/typescript.rs
MOmarMiraj: Added!

core/data/tests/test_byte_translation/output.ts
MOmarMiraj: Fixed!

core/src/language/python.rs
MOmarMiraj: Added!

core/src/language/go.rs
```diff
@@ -118,8 +118,19 @@ impl Language for Go {
         special_ty: &SpecialRustType,
         generic_types: &[String],
     ) -> Result<String, RustTypeFormatError> {
+        let mapped = if let Some(mapped) = self.type_map().get(&special_ty.get_nested_id()) {
+            mapped.to_owned()
+        } else {
+            String::new()
+        };
```
hculea, line 125: Can this already return `mapped` is it is not None?

core/src/language/go.rs
```diff
@@ -118,8 +118,19 @@ impl Language for Go {
...
         Ok(match special_ty {
-            SpecialRustType::Vec(rtype) => format!("[]{}", self.format_type(rtype, generic_types)?),
+            SpecialRustType::Vec(rtype) => {
+                // TODO: https://github.com/1Password/typeshare/issues/231
+                if rtype.contains_type(SpecialRustType::U8.id()) & !mapped.is_empty() {
+                    return Ok(mapped);
+                }
```
hculea, line 131: This shouldn't be required, right? That is, if we return `mapped` early.

core/src/language/python.rs
```diff
@@ -390,25 +444,66 @@ impl Python {
             .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
         let python_field_name = python_property_aware_rename(&field.id.original);
         let is_aliased = python_field_name != field.id.renamed;
-        match (not_optional_but_default, is_aliased) {
-            (true, true) => {
-                self.add_import("typing".to_string(), "Optional".to_string());
-                self.add_import("pydantic".to_string(), "Field".to_string());
+        // if more custom serialization/deserialization is needed, we can add it here and in the hashmap below
+        let requires_custom_translation = matches!(python_type.as_str(), "bytes");
```
hculea, line 448: instead of doing this here and having to hardcode "bytes", can we determine this based on the result of `get_custom_translation_function_names` which we can make return Option to suggest that some types don't require custom serialization/deserialization?

core/src/language/python.rs
```diff
@@ -390,25 +444,66 @@ impl Python {
...
+            is_aliased,
+            requires_custom_translation,
+        ) {
+            (true, true, false) => {
                 write!(w, "    {python_field_name}: Optional[{python_type}] = Field(alias=\"{renamed}\", default=None)", renamed=field.id.renamed)?;
             }
-            (true, false) => {
-                self.add_import("typing".to_string(), "Optional".to_string());
-                self.add_import("pydantic".to_string(), "Field".to_string());
+            (true, false, false) => writeln!(
```
hculea, line 459: This match statement grows exponentially - before we had 4 variants, now we have 8. Next time we add an option that may add a new type decorator, there will be 16.

However, a lot of these have things in common. 

For example, `{python_field_name}` is always the same.

Moreover, for types with custom serialization/deserialization, the decorated type simply looks like:
```
Annotated[<whatever the thing looks like without custom funcs>, BeforeValidator({deserialize_function}), PlainSerializer({serialize_function})]
```

I think we can break it down in 3 pieces:
* the field name - `{python_field_name}` is always the same
* the field type - start with the regular field type, wrap it with `Optional` if needed, wrap it with the validator and serializer if needed.
* The field default - add arguments `default=None` if it has a default `alias = `...` if it has an alias and construct it with `Field[<arguments>]`

We may end up with slightly more Rust code, but a lot less duplication between the Python one and in a much more maintainable state.

I'm happy to pair up on this if you'd like.

core/src/language/python.rs
```diff
@@ -700,6 +793,22 @@ fn handle_model_config(w: &mut dyn Write, python_module: &mut Python, fields: &[
     };
 }
 
+/// acquires custom translation function names if custom serialize/deserialize functions are needed
+fn get_custom_translation_function_names(python_type: &str) -> (String, String) {
```
hculea, line 797: nit - as per my comment above, we should make this return `Option<(String, String)>`

hculea: Let's also call this function more clearly - `json_translation_for_type` or something like that.

core/src/language/go.rs
MOmarMiraj: Yep, fixed!

core/src/language/python.rs
MOmarMiraj: Fixed!

core/src/language/python.rs
MOmarMiraj: Refactored to make it more readable.

core/src/language/typescript.rs
```diff
@@ -24,20 +24,55 @@ pub struct TypeScript {
...
     fn type_map(&mut self) -> &HashMap<String, String> {
         &self.type_mappings
     }
 
+    fn end_file(&mut self, w: &mut dyn Write) -> std::io::Result<()> {
+        if self.is_bytes {
+            return writeln!(
+                w,
+                r#"export function ReviverFunc(key: string, value: unknown): unknown {{
+    return Array.isArray(value) && value.every(v => Number.isFinite(v) && v >= 0 && v <= 255)
```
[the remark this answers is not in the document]
MOmarMiraj: Fixed!

core/src/language/python.rs
```diff
@@ -137,6 +160,7 @@ impl Language for Python {
         parameters: &[RustType],
         generic_types: &[String],
     ) -> Result<String, RustTypeFormatError> {
+        self.add_imports(base);
```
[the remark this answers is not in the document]
MOmarMiraj: This add common imports for packages like `Datetime` and `Url` which it wasn't before and fixed the testcase.

core/src/language/python.rs
```diff
@@ -83,12 +83,17 @@ pub struct Python {
     /// Whether or not to exclude the version header that normally appears at the top of generated code.
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
+    /// Whether or not to include the serialization/deserialzation functions for bytes.
+    /// This by default should be false as unless the user expclitly wants to translate to its bytes
+    /// representation
+    pub is_bytes: bool,
```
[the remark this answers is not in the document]
MOmarMiraj: Discussed with @hculea we will leave this pub as is due to the CLI and tests requiring to use `Default` and creating constructors just for this isn't worth it as this isn't used as a library a lot.

core/src/language/python.rs
MOmarMiraj: done!

core/src/language/go.rs
```diff
@@ -118,8 +118,14 @@ impl Language for Go {
...
     ) -> Result<String, RustTypeFormatError> {
+        if let Some(mapped) = self.type_map().get(&special_ty.to_string()) {
+            return Ok(mapped.to_owned());
+        };
+
         Ok(match special_ty {
-            SpecialRustType::Vec(rtype) => format!("[]{}", self.format_type(rtype, generic_types)?),
+            SpecialRustType::Vec(rtype) => {
+                format!("[]{}", self.format_type(rtype, generic_types)?)
+            }
```
hculea, line 128: Is this diff required?

MOmarMiraj: nope fixed!

core/src/language/python.rs
```diff
@@ -173,10 +193,23 @@ impl Language for Python {
...
+        }
         match special_ty {
-            SpecialRustType::Vec(rtype)
-            | SpecialRustType::Array(rtype, _)
-            | SpecialRustType::Slice(rtype) => {
+            SpecialRustType::Vec(rtype) => {
+                // TODO: https://github.com/1Password/typeshare/issues/231
+                self.add_import("typing".to_string(), "List".to_string());
+                Ok(format!("List[{}]", self.format_type(rtype, generic_types)?))
+            }
```
hculea, line 211: This diff is now no longer needed right? It can have the same translation as Array and Slice

core/src/language/python.rs
```diff
@@ -83,12 +91,17 @@ pub struct Python {
     /// Whether or not to exclude the version header that normally appears at the top of generated code.
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
+    /// Whether or not to include the serialization/deserialzation functions for bytes.
+    /// This by default should be false as unless the user expclitly wants to translate to its bytes
+    /// representation
```
hculea, line 96: This comment is no longer accurate

hculea (approved):

hculea:
Functionally tested and code reviewed, looks good! Thanks for the contribution here.

PR #233, opened 2025-02-26
Fix writing enum values in Go

edif2008:
## Summary

This PR fixes generating Algebraic Rust Enums in Go to properly set the constant values according to the serde configuration for that enum.

## Background

For Algebraic Rust Enums, when generating them in Go, the values of the constants were the original name of the enum variant instead of the renamed one, which takes into consideration the serde attributes. Because of that, it could happen that in such enums the JSON unmarshalling will fail due to not having any content to process. That was caused by the match case not hitting any variant.

### DIFF
--- a/core/data/tests/can_generate_algebraic_enum/output.go
+++ b/core/data/tests/can_generate_algebraic_enum/output.go
@@ -121,11 +121,11 @@ func NewAdvancedColorsTypeVariantReallyCoolType(content *ItemDetailsFieldValue) 
 type AdvancedColors2Types string
 const (
 	// This is a case comment
-	AdvancedColors2TypeVariantString AdvancedColors2Types = "String"
-	AdvancedColors2TypeVariantNumber AdvancedColors2Types = "Number"
-	AdvancedColors2TypeVariantNumberArray AdvancedColors2Types = "NumberArray"
+	AdvancedColors2TypeVariantString AdvancedColors2Types = "string"
+	AdvancedColors2TypeVariantNumber AdvancedColors2Types = "number"
+	AdvancedColors2TypeVariantNumberArray AdvancedColors2Types = "number-array"
 	// Comment on the last element
-	AdvancedColors2TypeVariantReallyCoolType AdvancedColors2Types = "ReallyCoolType"
+	AdvancedColors2TypeVariantReallyCoolType AdvancedColors2Types = "really-cool-type"
 )
 type AdvancedColors2 struct{ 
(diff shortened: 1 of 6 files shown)


core/src/language/go.rs
```diff
@@ -413,7 +413,7 @@ impl Go {
                         "\t{} {} = {:?}",
                         variant_type_const,
                         variant_key_type,
-                        &v.shared().id.original
+                        &v.shared().id.renamed
```
AndyTitu, line 416: What happens in the default case, where no serde renaming occurs? What casing is used then?

core/data/tests/can_generate_algebraic_enum/output.go
```diff
@@ -121,11 +121,11 @@ func NewAdvancedColorsTypeVariantReallyCoolType(content *ItemDetailsFieldValue)
 type AdvancedColors2Types string
 const (
 	// This is a case comment
-	AdvancedColors2TypeVariantString AdvancedColors2Types = "String"
-	AdvancedColors2TypeVariantNumber AdvancedColors2Types = "Number"
-	AdvancedColors2TypeVariantNumberArray AdvancedColors2Types = "NumberArray"
+	AdvancedColors2TypeVariantString AdvancedColors2Types = "string"
```
AndyTitu, line 124: Can you add a test for title case as well, just for completeness sake. I see everything was TitleCase before this change, so it makes sense to check that is still behaviour we can achieve somehow, especially since we might need to consider backwards compatibility for internal tools using this typeshare feature.

core/src/language/go.rs
edif2008: If no serde renaming occurs, it will match the original one, which is TItleCase.

core/data/tests/can_generate_algebraic_enum/output.go
edif2008: You mean a test case in which no serde renaming is used or a test in which renaming is title case? Or both?

AndyTitu: Where no renaming is used, as the most important.

edif2008: There's already a case in that test itself. Check `AdvancedColors` [here](https://github.com/1Password/typeshare/pull/233/files#diff-ec3820b150bb49f5f3b81093a7cf9d95954e5a433c2329dfb6203704fe8c73d7R9-R18). Notice how it wasn't changed as part of this PR.

AndyTitu (approved):

PR #234, opened 2025-02-28
Fix Bug in JS and Python Typesharing Custom Types

MOmarMiraj:
With #230 , we added custom translations for python and typescript which is used in the SDKs. Currently in main this will duplicate the serialization and deserialization for every type even if its the same. Also for the `deserialization` of bytes, we added an additional check to ensure that the length of the array is 0 so we don't convert empty arrays to the byte type.

### DIFF
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -9,11 +9,11 @@ def serialize_binary_data(value: bytes) -> list[int]:
 
 def deserialize_binary_data(value):
      if isinstance(value, list):
-         if all(isinstance(x, int) and 0 <= x <= 255 for x in value) & len(value) > 0:
-             return bytes(value)
+         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+            return bytes(value)
          raise ValueError("All elements must be integers in the range 0-255 (u8).")
      elif isinstance(value, bytes):
-             return value
+            return value
      raise TypeError("Content must be a list of integers (0-255) or bytes.")
 
 class Foo(BaseModel):
--- a/core/data/tests/test_byte_translation/output.py
+++ b/core/data/tests/test_byte_translation/output.py
@@ -9,7 +9,7 @@ def serialize_binary_data(value: bytes) -> list[int]:
 
 def deserialize_binary_data(value):
      if isinstance(value, list):
-         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+         if all(isinstance(x, int) and 0 <= x <= 255 for x in value) & len(value) > 0:
              return bytes(value)
          raise ValueError("All elements must be integers in the range 0-255 (u8).")
      elif isinstance(value, bytes):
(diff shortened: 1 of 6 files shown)


core/data/tests/test_byte_translation/output.py
```diff
@@ -9,7 +9,7 @@ def serialize_binary_data(value: bytes) -> list[int]:
 
 def deserialize_binary_data(value):
      if isinstance(value, list):
-         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+         if all(isinstance(x, int) and 0 <= x <= 255 for x in value) & len(value) > 0:
```
hculea, line 12: I don't think this (`& len(value) > 0:`) is required in Python, as here we are able to annotate only the types we want to transform with this. Let's remove this.

core/src/language/python.rs
```diff
@@ -92,7 +92,7 @@ pub struct Python {
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
     /// Carries the content of the custom serializer and deserializer functions if required
-    pub custom_json_translation_functions: Vec<String>,
+    pub custom_json_translation_functions: HashSet<String>,
```
hculea, line 95: This is not a proper naming for this type.

Let's call it something like `types_for_custom_json_translation`

core/src/language/typescript.rs
```diff
@@ -25,7 +26,7 @@ pub struct TypeScript {
     /// If you aren't generating a snapshot test, this setting can just be left as a default (false)
     pub no_version_header: bool,
     /// Carries the content of the custom reviver/replacer content if needed.
-    pub custom_json_translation_functions: Vec<CustomJsonTranslationContent>,
+    pub custom_json_translation_functions: HashSet<String>,
```
hculea, line 29: same for this

core/src/language/python.rs
```diff
@@ -138,12 +138,24 @@ impl Language for Python {
 
         self.write_all_imports(w)?;
 
-        self.custom_json_translation_functions.iter().try_for_each(
-            |custom_translation_function| -> std::io::Result<()> {
-                writeln!(w, "{custom_translation_function}")?;
-                writeln!(w)
-            },
-        )?;
+        if !self.custom_json_translation_functions.is_empty() {
```
hculea, line 141: This check is not required - if empty, the iter would simply not iterate

core/data/tests/test_byte_translation/output.py
MOmarMiraj: Ahh i see okay removed!

core/src/language/python.rs
MOmarMiraj: Fixed! and updated comment

core/src/language/typescript.rs
MOmarMiraj: Fixed! and updated comment

core/src/language/python.rs
MOmarMiraj: Removed!

core/src/language/typescript.rs
```diff
@@ -40,7 +41,12 @@ impl Language for TypeScript {
     }
 
     fn end_file(&mut self, w: &mut dyn Write) -> std::io::Result<()> {
-        if !self.custom_json_translation_functions.is_empty() {
+        if !self.types_for_custom_json_translation.is_empty() {
```
hculea, line 44: this `.is_empty()` check is not required

hculea (approved):
Approving the code, have not functionally tested. Please do load these in the SDKs to validate functionally before merging.

PR #236, opened 2025-03-12
Add support for DateTime field for Go, JS, Python 

MOmarMiraj:
This PR will add `DateTime` support from the rust `std::time::OffSetTime` library for Go, JS, and Python. The other languages are not supported yet as lack of expertise in those areas.

I am keeping the same implementation for Scala, Kotlin, and Swift to ensure backwards compatibility. (i.e. if you used to type map DateTime -> string).


core/src/rust_types.rs
```diff
@@ -578,6 +584,7 @@ impl SpecialRustType {
             Self::Slice(_) => "&[]",
             Self::Option(_) => "Option",
             Self::HashMap(_, _) => "HashMap",
+            Self::DateTime => "DateTime",
```
MOmarMiraj, line 587: question for reviewer:

Should this be DateTime or OffsetDateTime?

core/src/rust_types.rs
```diff
@@ -369,6 +371,7 @@ impl TryFrom<&syn::Type> for RustType {
                             params.next().unwrap().into(),
                         ))
                     }
+                    "OffsetDateTime" => Self::Special(SpecialRustType::DateTime),
```
hculea, line 374: this is pretty unlinkely, but this would represent a breaking change in the off-chance that someone actually named their structure, this. 

can we only parse the full `time::OffsetDateTime` type path for now?

core/data/tests/test_datetime_translation/output.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+from datetime import datetime
+from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSerializer
+from typing import Annotated
+
+
+def serialize_datetime_data(utc_time: datetime) -> str:
+        return utc_time.strftime("%Y-%m-%dT%H:%M:%SZ")
```
hculea, line 9: once concern that I have here is that there are multiple zulu formats available:
* YYYY-MM-DDTHH:MM:SSZ → 2025-03-13T15:30:45Z (Date + Time + Seconds)
* YYYY-MM-DDTHH:MM:SS.sssZ → 2025-03-13T15:30:45.123Z (Including milliseconds)
* YYYY-MM-DDTHH:MM:SS.ssssssZ → 2025-03-13T15:30:45.123456Z (Including microseconds)
* YYYY-MM-DDTHH:MM:SS.sssssssssZ → 2025-03-13T15:30:45.123456789Z (Including nanoseconds)

and I am wondering if the presumed serialization method (rfc3339) consistently produces this format. Can we check and account for the other formats, otherwise?

core/data/tests/test_datetime_translation/output.py
```diff
@@ -0,0 +1,21 @@
...
+
+def deserialize_datetime_data(utc_time_str: str) -> datetime:
+        return datetime.strptime(utc_time_str, "%Y-%m-%dT%H:%M:%SZ")
```
hculea, line 12: here I think we can actually default to the milliseconds granularity instead? 

This would help with the use-cases where precision is required and should still be rfc3339 de-serializable.

core/data/tests/test_byte_translation/output.ts
```diff
@@ -11,14 +11,14 @@ export interface Foo {
  */
 export const ReviverFunc = (key: string, value: unknown): unknown => {
```
hculea, line 12: can we add an example where we have both bytes and times to get more confidence on how this file would look like?

core/src/language/typescript.rs
```diff
@@ -81,9 +80,9 @@ export const ReplacerFunc = (key: string, value: unknown): unknown => {{
         generic_types: &[String],
     ) -> Result<String, RustTypeFormatError> {
         if let Some(mapped) = self.type_mappings.get(&special_ty.to_string()) {
-            if custom_translations(mapped).is_some() {
+            if self.custom_translations(mapped).is_some() {
                 self.types_for_custom_json_translation
-                    .insert(mapped.to_string());
+                    .insert(mapped.to_string(), Vec::new());
             }
```
hculea, line 86: We're using Vec::new() and `is_empty()` as a placeholder for None. I think we should model the value as an Option to capture this better in the data structure too.

core/src/rust_types.rs
```diff
@@ -249,6 +249,8 @@ pub enum SpecialRustType {
     HashMap(Box<RustType>, Box<RustType>),
     /// Represents `Option<T>` from the standard library
     Option(Box<RustType>),
+    /// Represents time::OffsetDateTime from time
+    DateTime,
```
hculea, line 253: Can we also document here the expected serialization/deserialization? Let's also link to the serde function that we expect to be used.

core/src/language/typescript.rs
MOmarMiraj: Discussed with @hculea and decided to model it as is due to the convolution of making this into an `Option<Vec<T>>`

core/src/rust_types.rs
MOmarMiraj: Discussed with @hculea, this would require a potential breaking change in `try_from` so decided to opt against this.

core/src/language/go.rs
```diff
@@ -177,7 +178,13 @@ impl Language for Go {
         }
         writeln!(w, "package {}", self.package)?;
         writeln!(w)?;
-        writeln!(w, "import \"encoding/json\"")?;
+        writeln!(
+            w,
+            r#"import (
+    "encoding/json"
+    "time"
```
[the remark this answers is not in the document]
MOmarMiraj: Done, added capability to import at will for Go, took a similar approach to python.

core/src/rust_types.rs
MOmarMiraj: Added!

core/data/tests/test_byte_translation/output.ts
MOmarMiraj: Removed the `test_byte_translation` and consolidated both tests

core/data/tests/test_datetime_translation/output.py
MOmarMiraj: Done!

core/data/tests/test_custom_serialize_deserialize_functions/output.go
```diff
@@ -0,0 +1,14 @@
+package proto
+
+import (
+    "encoding/json,
+	time"
```
hculea, line 5: This should be:
```suggestion
    "encoding/json"
    "time"
```

core/src/rust_types.rs
MOmarMiraj: `OffsetDateTime` is the more correct option and was changed.

core/data/tests/anonymous_struct_with_rename/output.go
```diff
@@ -1,6 +1,8 @@
 package proto
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
```
hculea, line 5: note to reduce this diff:

when writing the imports, you can match based on the number of imports:
* if 0, do nothing
* if 1, do `import "your-import"`
* if more, make a block like you are doing here.

MOmarMiraj: done!

core/data/tests/test_custom_serialize_deserialize_functions/output.ts
```diff
@@ -10,14 +13,20 @@ export interface Foo {
...
     }
     return value;
 };
 
 export const ReplacerFunc = (key: string, value: unknown): unknown => {
-    if (value instanceof Uint8Array) {
+    if (value instanceof Date) {
+        return value.toISOString();
+    }
+	if (value instanceof Uint8Array) {
```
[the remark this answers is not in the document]
MOmarMiraj: fixed to be spaces!

core/data/tests/test_custom_serialize_deserialize_functions/output.ts
```diff
@@ -10,14 +13,20 @@ export interface Foo {
  * These functions allow for flexible encoding and decoding of data, ensuring that complex types are properly handled when converting between TS objects and JSON
  */
 export const ReviverFunc = (key: string, value: unknown): unknown => {
-    if (Array.isArray(value) && value.every(v => Number.isInteger(v) && v >= 0 && v <= 255) && value.length > 0)  {
+    if (/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$/.test(value as string) && (key == "time" || key == "time2" || key == "time3")) {
```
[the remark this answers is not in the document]
MOmarMiraj: Added!

core/src/language/go.rs
```diff
@@ -536,6 +542,27 @@ func ({short_name} {full_name}) MarshalJSON() ([]byte, error) {{
         };
         self.acronyms_to_uppercase(&name)
     }
+
+    fn write_import(&mut self, name: &str) {
```
hculea, line 546: this does not write anything, so it should be called `add_import`

core/src/language/typescript.rs
```diff
@@ -402,6 +391,40 @@ impl TypeScript {
         }
         Ok(())
     }
+
+    fn custom_translations(&self, ts_type: &str) -> Option<CustomJsonTranslationContent> {
+        let id = self
+            .types_for_custom_json_translation
+            .get(ts_type)
+            .filter(|ids| !ids.is_empty())
+            .map(|ids| format!(" && (key == \"{}\")", ids.join("\" || key == \"")));
```
hculea, line 400: I don't really understand this, can't this be something like:
```
let condition = format!("&& ({})", 
  ids.iter().map(|id| format!("key == {}", id)).join(" || ")
)
```
?

core/data/tests/test_custom_serialize_deserialize_functions/output.ts
```diff
@@ -10,13 +13,19 @@ export interface Foo {
  * These functions allow for flexible encoding and decoding of data, ensuring that complex types are properly handled when converting between TS objects and JSON
  */
 export const ReviverFunc = (key: string, value: unknown): unknown => {
+    if (/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?Z$/.test(value as string) && (key == "time" || key == "time2" || key == "time3")) {
```
hculea, line 16: can we also test that value is a string, as we are checking below that it is an array?

hculea: we should do this BEFORE testing the regex and doing "as string"

hculea: maybe with something like "typeof value === "string""

core/data/tests/test_custom_serialize_deserialize_functions/input.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub bytes: Vec<u8>,
+    pub bytes2: Vec<u8>
+}
+
+#[typeshare]
+#[serde(rename_all = "camelCase")]
+pub struct TwoFoo {
+    pub time: time::OffsetDateTime,
+    pub bytes: Vec<u8>,
+}
```
hculea, line 16: nit: new line EOF

core/data/tests/test_custom_serialize_deserialize_functions/output.py
```diff
@@ -0,0 +1,49 @@
...
+     if isinstance(value, list):
+         if all(isinstance(x, int) and 0 <= x <= 255 for x in value):
+            return bytes(value)
+         raise ValueError("All elements must be integers in the range 0-255 (u8).")
+     elif isinstance(value, bytes):
+            return value
+     raise TypeError("Content must be a list of integers (0-255) or bytes.")
+
+def serialize_datetime_data(utc_time: datetime) -> str:
+        return utc_time.strftime("%Y-%m-%dT%H:%M:%SZ")
```
hculea, line 21: I believe we discussed to make this "%Y-%m-%dT%H:%M:%S.%fZ" right?

core/src/language/go.rs
```diff
@@ -18,6 +18,8 @@ use super::CrateTypes;
 pub struct Go {
     /// Name of the Go package.
     pub package: String,
+    /// HashMap<ModuleName, HashSet<Identifier>
+    pub imports: HashSet<String>,
```
hculea, line 22: this will also not be deterministic, right? Should this be a BTreeMap?

hculea (approved):

core/data/tests/test_custom_serialize_deserialize_functions/output.py
MOmarMiraj: Yep, fixed this!

core/src/language/go.rs
MOmarMiraj: Yep and I fixed the comment to better reflect the type aswell.

PR #238, opened 2025-03-17
docs: Java RFC

bcheidemann:
The contributing docs suggest to open an RFC issue "in order to discuss the broader impacts of the proposed change with the other maintainers". I think this is an excellent suggestion, but I felt that since this RFC touches on quite a few important aspects of the Typeshare implementation, it may be sensible to track it as a PR, so maintainers can leave comments on specific aspects. If an issue is preferred though, I will happily open one instead 🙂

Looking forward to feedback! ❤️

### DIFF
--- a/docs/src/rfcs/001-java-language-support/README.md
+++ b/docs/src/rfcs/001-java-language-support/README.md
@@ -1,0 +1,504 @@
+# Summary
+
+Typeshare currently supports several popular languages, but lacks support for Java. Adding support for Java would improve the state of interoperability between Rust and Java codebases. However, Java has several constraints not present in the currently supported languages. This RFC outlines a proposal for how Java support can be implemented, taking into consideration the languages idiosyncrasies.
+
+# Basic example
+
+This input:
+
+```rust
+/// This is a comment.
+#[typeshare]
+pub struct Person {
+    pub name: String,
+    pub age: u8,
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/rfcs/001-java-language-support/README.md
```diff
@@ -0,0 +1,505 @@
...
+}
+```
+
+```java
+record MaybePerson(
+	Person person
+) { }
+```
+
+## Single File per Class & File Naming
```
bcheidemann, line 456: Having had some time to think about this, a simpler solution could be to use a sort of namespace class to group classes/records. For example, if generating types for the crate "lib_shared_types", this could look something like:

```java
class LibSharedTypes {
    public static record Person(
        String name,
        int age
    ) {}
}
```

Which can be used like this:

```java
new LibSharedTypes.Person("...", 42);
```

Classes can reference each other as normal within the same "namespace":

```java
class LibSharedTypes {
    public static record Person(
        String name,
        short age
    ) {}

    public static record Couple(
        Person person1,
        Person person2
    ) {}
}
```

Which can then be used like this:

```java
var bill = new LibSharedTypes.Person("Bill", 73);
var ben = new LibSharedTypes.Person("Ben", 73);
var flowerPotMen = new LibSharedTypes.Couple(bill, ben);
```

PR #241, opened 2025-03-21
Update typescript.rs

harshils23:
replaced 2 = with 3= to properly check equality

### DIFF
--- a/core/src/language/typescript.rs
+++ b/core/src/language/typescript.rs
@@ -403,7 +403,7 @@ impl TypeScript {
             .map(|ids| {
                 format!(
                     " && ({})",
-                    ids.iter().map(|id| format!("key == \"{id}\"")).join(" || ")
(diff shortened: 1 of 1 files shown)


MOmarMiraj (approved):
LGTM!

core/src/language/typescript.rs
```diff
@@ -403,7 +403,7 @@ impl TypeScript {
             .map(|ids| {
                 format!(
                     " && ({})",
-                    ids.iter().map(|id| format!("key == \"{id}\"")).join(" || ")
+                    ids.iter().map(|id| format!("key === \"{id}\"")).join(" || ")
```
MOmarMiraj, line 406: The test pipeline is failing due to you not updating the tests.

Take a look here https://github.com/1Password/typeshare/blob/main/core/README.md#testing on how to update the tests specifically you would need to run the `env UPDATE_EXPECT=1 cargo test -p typeshare-core` command to update the tests. As well as format the code using `cargo fmt --all`

PR #242, opened 2025-03-21
Fix Equality Check Bug in typescript generator file

### LINKED ISSUE #240: Change == to === for JS Reviver Functions
When comparing the key to the field name inside of `core/src/language/typescript.rs` inside the `custom_translation` function, we should use the triple equal signs according to eslint. The error is explained more [here](https://eslint.org/docs/latest/rules/eqeqeq)

harshils23:
made 2 = to 3 = for correct equality check

Resolves #240


MOmarMiraj (approved):
LGTM!, Please ensure your commits are verified as your unable to merge unless its verified.

PR #245, opened 2025-04-08
Fixed conditional writing of Codable.swift

chadseld:
Moved location of newline character generation so that the file contents diff check inside write_codable_file will work as expected.

The Codable.swift file in only supposed to be written to disk if it has changed. Currently, the Codeable.swift file is always being written to disk. This causes problems with Xcode previews. When the typeshare script is run as part of an Xcode preview build, Xcode detects a file change (Codeable.swift) and restarts the compilation in an endless loop.

A previous fix to this bug was made/attempted here: https://github.com/1Password/typeshare/pull/205.

### DIFF
--- a/core/src/language/swift.rs
+++ b/core/src/language/swift.rs
@@ -790,12 +790,12 @@ impl Swift {
             decs.push(CODABLE);
         }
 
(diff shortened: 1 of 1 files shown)


chadseld:
Should the Clippy errors be fixed on a separate PR? They are not due to changes made here.

PR #251, opened 2025-05-31
add test with type using #[unsafe(no_mangle)].

darrell-roberts:
Test to assert typeshare works with `#[unsafe(no_mangle)]` annotation.


charlespierce (approved):
LGTM!

PR #252, opened 2025-05-31
Update for newer clippy lints

darrell-roberts:
Updated clippy lints.


Lucretiel:
Assuming all the semver-major dependency updates are good, this looks good to me

MOmarMiraj (approved):
these clippy lints look good and pipelines giving us confidence that nothing has broken Approved!

PR #253, opened 2025-06-09
Add support for Kotlin custom serializers

vrendina:
When Kotlin classes are marked with `@Serializable` the Kotlinx serialization compiler plugin automatically generates an implementation of the `KSerializer` interface for the annotated class. In most cases this is fine, but we have run into a scenario where a Rust enum type has significant number of entries (1200+) and the generated serializer is exceeding the JVM maximum method size limit. We need to be able to specify a custom serializer that uses a different strategy to look up serializers for each entry in the enum or the application will fail to compile. 

See https://github.com/Kotlin/kotlinx.serialization/issues/3017

We have a Rust enum that is generating a sealed class hierarchy that looks approximately like the example below. 

```kotlin
@Serializable
sealed class GeneratedSealedClassExample {
    @Serializable
    data object GeneratedSealedClassExample1: GeneratedSealedClassExample()

    ... 

    @Serializable
    data object GeneratedSealedClassExample2500: GeneratedSealedClassExample()
}
```

When the KotlinX serialization compiler plugin runs it generates a serializer that has two arrays containing the class types and serializers for every subclass. As this continues to grow it will exceed the maximum method size and we will not longer be able compile. 

```kotlin
@NotNull
private static final Lazy<KSerializer<Object>> $cachedSerializer$delegate = LazyKt.lazy(LazyThreadSafetyMode.PUBLICATION, () -> {
    return new SealedClassSerializer("org.example.GeneratedSealedClassExample",
            Reflection.getOrCreateKotlinClass(GeneratedSealedClassExample.class),
            new KClass[]{
                    Reflection.getOrCreateKotlinClass(GeneratedSealedClassExample1.class),
                    ...
                    Reflection.getOrCreateKotlinClass(GeneratedSealedClassExample2500.class)
            },
            new KSerializer[]{
                    new ObjectSerializer("org.example.GeneratedSealedClassExample.GeneratedSealedClassExample1", GeneratedSealedClassExample1.INSTANCE, new Annotation[0]),
                    ...
                    new ObjectSerializer("org.example.GeneratedSealedClassExample.GeneratedSealedClassExample2500", GeneratedSealedClassExample2500.INSTANCE, new Annotation[0])
            },
            new Annotation[0]
    );
});
```

Our solution is to utilize a custom serializer that uses runtime reflection to look up the necessary implementation of the `KSerializer`. This solution is approximately 50x slower than the current generated implementation, so ultimately we may need to find another way to look up serializers if that becomes problematic. Benchmarking shows this is on the order of approximately 25 microseconds which is acceptable.

### DIFF
--- a/app/engine/src/args.rs
+++ b/app/engine/src/args.rs
@@ -133,7 +133,7 @@ pub fn add_personalizations(
     command: clap::Command,
     personalizations: PersonalizeClap,
 ) -> clap::Command {
-    let command = command.arg(
+    let command = command.disable_version_flag(true).arg(
         clap::Arg::new("version")
             .short('V')
             .long("version")
--- a/app/engine/src/args.rs
+++ b/app/engine/src/args.rs
@@ -148,12 +148,10 @@ pub fn add_personalizations(
         None => command,
     };
 
-    let command = match personalizations.about {
+    match personalizations.about {
         Some(about) => command.about(about),
         None => command,
-    };
-
-    command
+    }
 }
 
 /// Add a `--lang` argument to the command. This argument will be optional if
--- a/app/engine/src/args.rs
+++ b/app/engine/src/args.rs
@@ -133,13 +133,6 @@ pub fn add_personalizations(
     command: clap::Command,
     personalizations: PersonalizeClap,
 ) -> clap::Command {
-    let command = command.disable_version_flag(true).arg(
-        clap::Arg::new("version")
-            .short('V')
-            .long("version")
-            .action(clap::ArgAction::Version),
-    );
-
     let command = match personalizations.name {
         Some(name) => command.name(name),
         None => command,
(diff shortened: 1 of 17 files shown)


app/engine/src/args.rs
```diff
@@ -133,7 +133,7 @@ pub fn add_personalizations(
     command: clap::Command,
     personalizations: PersonalizeClap,
 ) -> clap::Command {
-    let command = command.arg(
+    let command = command.disable_version_flag(true).arg(
```
vrendina, line 136: I was not able to run Typeshare with this missing. 

```
❯ ./target/debug/typeshare2

thread 'main' panicked at /Users/victor.rendina/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.35/src/builder/debug_asserts.rs:86:13:
Command typeshare2-cli: Argument names must be unique, but 'version' is in use by more than one argument or group (call `cmd.disable_version_flag(true)` to remove the auto-generated `--version`)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Issue #260, opened 2025-08-01
Provide more helpful error messages for


stewartadam:
Some of the error messages don't provide any line numbers of the originating struct name, making them annoying to pin down:

```
[2025-07-31 23:21:48.594360 -07:00] ERROR [/Users/stewartadam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/typeshare-cli-1.13.3/src/main.rs:293] Parsing error: "failed to parse a rust type: Tuples are not allowed in typeshare types" in file "../../git/project/src/outbound.rs"

[2025-07-31 23:21:20.080729 -07:00] ERROR [/Users/stewartadam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/typeshare-cli-1.13.3/src/main.rs:293] Parsing error: "failed to parse a rust type: ["u64"]" in file "../../git/project/src/command_types.rs"
```

Would be nice to provide more helpful error messages that include the struct/enum name, like some of them do:

```
[2025-07-31 23:21:20.080748 -07:00] ERROR [/Users/stewartadam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/typeshare-cli-1.13.3/src/main.rs:293] Parsing error: "serde tag attribute needs to be specified for algebraic enum Attribute. e.g. #[serde(tag = "type", content = "content")]" in file "../../git/project/src/attributes.rs"
```

stewartadam:
Similarly, when a type parse failure occurs from a `#[typeshare(serialized_as =  "...")]` the error is opaque and doesn't reference the type that failed to parse:

```
Parsing error: "failed to parse a rust type: []" in file "crates/src/outbound.rs"
```

darrell-roberts:
Created an MR that adds line and column number reporting to parser failure/rejections. 

Example test invalid type parse error:

```
cat /tmp/typeshare-test/src/test.rs
//! This is a module comment

#[typeshare]
pub struct MyType {
    pub field: usize
}
```

Running from the command line.
```
cargo run --bin typeshare -- -l typescript -d /tmp/out /tmp/typeshare-test
   Compiling typeshare-core v1.13.3 (/home/droberts/projects/typeshare/core)
   Compiling typeshare-cli v1.13.3 (/home/droberts/projects/typeshare/cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.84s
     Running `target/debug/typeshare -l typescript -d /tmp/out /tmp/typeshare-test`
[...]

PR #262, opened 2025-08-02
Use Span to capture line and column numbers for parsing rejections.

darrell-roberts:
It is useful to capture the line and column numbers when parsing rejects the Rust source code. 

This MR uses `Span` to capture this data. The error handling could be reworked some more but this allows capturing the line, column numbers without too many changes.

This was requested in #260 

Example test invalid type parse error:

```
cat /tmp/typeshare-test/src/test.rs
//! This is a module comment

#[typeshare]
pub struct MyType {
    pub field: usize
}
```

Running from the command line.
```
cargo run --bin typeshare -- -l typescript -d /tmp/out /tmp/typeshare-test
   Compiling typeshare-core v1.13.3 (/home/droberts/projects/typeshare/core)
   Compiling typeshare-cli v1.13.3 (/home/droberts/projects/typeshare/cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.84s
     Running `target/debug/typeshare -l typescript -d /tmp/out /tmp/typeshare-test`
[2025-08-02 17:25:09.911115 -04:00] INFO [cli/src/main.rs:78] typeshare started generating types
[2025-08-02 17:25:09.911217 -04:00] INFO [cli/src/main.rs:85] Using directories: ["/tmp/typeshare-test"]
[2025-08-02 17:25:09.918404 -04:00] ERROR [cli/src/main.rs:296] Parsing error: "Failed to parse a Rust type: Unsupported type: "usize", on line 5 and column 15" in file "/tmp/typeshare-test/src/test.rs"
[2025-08-02 17:25:09.918466 -04:00] ERROR [cli/src/main.rs:304] Errors encountered during parsing.
[2025-08-02 17:25:09.918505 -04:00] ERROR [cli/src/main.rs:72] typeshare failed to generate types: Errors encountered during parsing.
Error: Errors encountered during parsing.
```

Notable parse error with file, line and column number:

Parsing error: "Failed to parse a Rust type: Unsupported type: "usize", on line 5 and column 15" in file "/tmp/typeshare-test/src/test.rs"

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -1,8 +1,7 @@
 //! Error types for parsing.
+use crate::rust_types::RustTypeParseError;
 use proc_macro2::Span;
 use thiserror::Error;
-
-use crate::rust_types::RustTypeParseError;
 
 #[derive(Debug)]
 /// Wrapper for a parse error which includes a span.
@@ -61,10 +60,28 @@ pub enum ParseError {
     IOError(String),
 }
 
-/// Convert a [`RustTypeParseError`] into a [`ParseErrorWithSpan`].
-pub fn rust_type_parse_err(err: RustTypeParseError, span: Span) -> ParseErrorWithSpan {
-    ParseErrorWithSpan {
-        error: ParseError::RustTypeParseError(err),
-        span,
+impl RustTypeParseError {
+    /// Convert a [`RustTypeParseError`] into a [`ParseErrorWithSpan`].
+    pub fn with_span(self, span: Span) -> ParseErrorWithSpan {
+        ParseError::RustTypeParseError(self).with_span(span)
+    }
+}
+
+impl ParseError {
+    /// If a condition is not met then call the error function.
+    pub fn ensure(
+        cond: bool,
+        mut f: impl FnMut() -> ParseErrorWithSpan,
+    ) -> Result<(), ParseErrorWithSpan> {
+        if !cond {
+            Err(f())
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Convert a [`ParseError`] into a [`ParseErrorWithSpan`].
+    pub fn with_span(self, span: Span) -> ParseErrorWithSpan {
+        ParseErrorWithSpan { error: self, span }
     }
 }
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -1,0 +1,65 @@
+use proc_macro2::Span;
+use thiserror::Error;
+
+use crate::rust_types::RustTypeParseError;
+
+#[derive(Debug)]
+pub struct ParseErrorWithSpan {
+    pub error: ParseError,
+    pub span: Span,
+}
+
+impl std::error::Error for ParseErrorWithSpan {}
+
+impl std::fmt::Display for ParseErrorWithSpan {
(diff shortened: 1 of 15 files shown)


core/src/error.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn ensure(
+        cond: bool,
+        mut f: impl FnMut() -> ParseErrorWithSpan,
+    ) -> Result<(), ParseErrorWithSpan> {
+        if !cond {
+            Err(f())
+        } else {
+            Ok(())
+        }
+    }
```
charlespierce, line 81: From an API design standpoint, it feels a little odd to have an associated function on `ParseError` that actually returns / works with `ParseErrorWithSpan` types and doesn't work directly with `ParseError` at all.

Would it break things to have this work with `ParseError` directly? For ergonomic reasons, I could see adding a `WithSpan` helper trait and implementing it on `Result<T, ParseError>` to enable easy conversion to `Result<T, ParseErrorWithSpan>`

Alternatively, since this is taking in a `bool` directly (as opposed to a predicate)—is it a lot more code to have callers use an `if` statement to build their own `Result` values?

core/src/parser.rs
```diff
@@ -50,47 +50,13 @@ impl DecoratorKind {
...
-}
-
 /// Error with it's related data.
 #[derive(Debug)]
 pub struct ErrorInfo {
     /// The file name being parsed.
     pub file_name: String,
     /// The parse error.
-    pub error: ParseError,
+    pub error: String,
```
charlespierce, line 59: Out of curiosity, why was this changed to a `String` instead of a typed error?

core/src/rust_types.rs
```diff
@@ -314,7 +317,7 @@ impl Display for SpecialRustType {
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
 pub enum RustTypeParseError {
-    #[error("{0:?}")]
+    #[error("Unsupported type: \"{}\"", .0.iter().join(","))]
```
charlespierce, line 320: Small organization question: Would it be confusing to move this into `error.rs` along side the other parse errors? It seems the `with_span` method is defined there, so it feels like it might be worthwhile to have this there as well for consistency.

core/tests/agnostic_tests.rs
```diff
@@ -1,35 +1,34 @@
...
 };
+
 /// Parse and generate types for a single Rust input file.
 pub fn process_input(
     input: &str,
     language: &mut dyn Language,
     imports: &CrateTypes,
     out: &mut dyn Write,
-) -> Result<(), ProcessInputError> {
+) -> anyhow::Result<()> {
```
charlespierce, line 15: For my own understanding (similar to the question about using `String` above), what benefits does this have over the existing approach of returning a typed error?

charlespierce (changes requested):
Overall looks great, and the improved error messages are fantastic! A few code organization questions, but nothing that would impact the main approach.

core/src/error.rs
darrell-roberts: Yeah, `ensure` here does not hold much value. Added the `WithSpan` trait and updated accordingly.

PR #263, opened 2025-08-08
Bump nix flake.lock

LuminaSapphira:


savannidgerinel:
The changes look good and Typeshare still launches correctly:

image

PR #264, opened 2025-08-13
Add automated job to update flake.lock file

charlespierce:
Info
-----
* The example `flake.lock` file we had was out of date and causing potential issues.
* That was manually fixed in #263, however to prevent it getting out of date again, we should automate any future updates.
* The action https://github.com/DeterminateSystems/update-flake-lock provides exactly what we are looking for - automated creation of PRs to bump `flake.lock`

Changes
-----
* Used the example from `update-flake-lock` to create an automated job that will create a PR whenever there are updates to `flake.lock` necessary

### DIFF
--- a/.github/workflows/update_flake_lock.yml
+++ b/.github/workflows/update_flake_lock.yml
@@ -1,0 +1,19 @@
+name: "Check for updates to Nix flake.lock file"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 0" # runs weekly on Sunday at midnight
+
+jobs:
+  nix-flake-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/determinate-nix-action@v3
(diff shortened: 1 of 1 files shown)


savannidgerinel:
I have some concerns about it not creating a new PR in this case, and the messages in the log don't seem to apply here:

image

I'll approve the MR, but we should immediately run the job to ensure that it will generate a flake.lock update. Nix unstable updates frequently, so we should get a flake.lock update every time we run this job.

charlespierce:
I think the logs about not running jobs is that for a PR created by an existing job, it won't automatically run CI on it. I assume that's so creating a job that runs on opening a PR and opens its own new PR doesn't immediately cause an infinite loop of CI.

Looking at the logs for the end of the job, it looks like it detected no changes and so it didn't create a PR:
Screenshot 2025-08-13 at 11 03 31 AM

That said, if Nix updates frequently enough that we expect there to be changes, I'll take a look and manually run the commands locally to verify that there actually isn't anything before I merge to avoid churn. Thanks for the review!

PR #267, opened 2025-08-26
Add automated job to update flake.lock file with signed commits

charlespierce:
Info
-----
* Second attempt at an automated job to update `flake.lock`, after #264
* The original job worked, but the PR created did not have signed commits, so it wasn't mergeable.
* This approach adds a GPG key and associates the commits with @1password-bot (which is already used in other repos within the org)

Changes
-----
* The job is created the same way as the previous PR, however we now add the author, committer, and GPG key information using repo secrets.

Testing
-----
* We discovered in the last PR that testing via a PR-triggered job did not work as expected, so we'll likely need some quick follow-up testing once this is merged.

### DIFF
--- a/.github/workflows/update_flake_lock.yml
+++ b/.github/workflows/update_flake_lock.yml
@@ -1,0 +1,26 @@
+name: "Check for updates to Nix flake.lock file"
+
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


LuminaSapphira (approved):
Hopefully this one works!

charlespierce:
Success! https://github.com/1Password/typeshare/pull/268 (and even merged)

